PR #4, opened 2015-07-17
MQTT: fix for incorrect UNSUBSCRIBE decoding / encoding

### LINKED ISSUE #5: MqttDecoder and MqttEncoder fail to process UNSUBSCRIBE packets
* `MqttDecoder` does not expect QoS 1 value on UNSUBSCRIBE packet's first byte (which is the only correct QoS value for this packet type)
* `MqttEncoder` counts 1 extra byte per topic filter while calculating remaining length value.

nayato:
motivation: Fix #5.

Also, adding convenient overload ctors on packets.

### DIFF
--- a/src/DotNetty.Codecs.Mqtt/MqttEncoder.cs
+++ b/src/DotNetty.Codecs.Mqtt/MqttEncoder.cs
@@ -335,7 +335,7 @@ namespace DotNetty.Codecs.Mqtt
                 foreach (string topic in packet.TopicFilters)
                 {
                     byte[] topicFilterBytes = EncodeStringInUtf8(topic);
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
IMHO: create a test specifically for this bug. We do it out of habit on Helios / Akka.NET when we come across remoting-edge cases like this. Call it `Verify_Issue_5_fixed()` or something like that.

Wouldn't do it for every single bug, but for ones that deal with specific wire formats and protocols (end to end stuff) I think it's a good idea.

nayato:
added a case in e2e test for now. It isn't an edge case, it's basic functionality that hasn't been tested before. Created #6 for that.

PR #12, opened 2015-08-07
ChannelHandlerAdapter: virtual where missing + impl deregister

nayato:
Correctly placed "virtual", implemented default DeregisterAsync behavior.
Minor tweaks to ReSharper settings.


Aaronontheweb:
Looks good to me. I'd merge it.

src/DotNetty.Transport/Channels/IChannelHandler.cs
```diff
@@ -89,6 +89,7 @@ public interface IChannelHandler
         [PipelinePropagation(PropagationDirections.Inbound)]
         void ExceptionCaught(IChannelHandlerContext context, Exception exception);
 
+        [PipelinePropagation(PropagationDirections.Outbound)]
```
Aaronontheweb: One nitpick - the `PipelinePropagation` and `Skip` attributes need some XML-DOC comments that explain what they do inside PipelinePropagationAttribute.cs and SkipAttribute.cs respectively. Not immediately clear to me :p

PR #13, opened 2015-08-10
DotNetty Build System

Aaronontheweb:
We're using FAKE to create builds, just like Akka.NET, Helios, and oodles of other projects.

The key files to pay attention to are as follows:
1. `build.cmd` and `build.sh` - this is how you launch FAKE on Windows and Linux repsectively. It runs some NuGet.exe commands to make sure that the test runners and build system binaries are available before the execution begins.
2. `build.fsx` - this is the F# script that actually defines the build system.
3. all of the individual `.nuspec` files for each project - FAKE scans for these to determine which assemblies will be deployed in a NuGet package and which ones won't.
4. `src\SharedAssemblyInfo.cs` - linked by all NuGet-deployed binaries and auto-generated by FAKE. This is used to synchronize the major parts of the build number by what's inside `RELEASE_NOTES.md`, which has to be authored by a human. The build number itself is determined by the build count on TeamCity.
5. `RELEASE_NOTES.md` - this gets parsed into the release notes for each NuGet package and determines what the version number of this release is.

To see all of the different options for running builds, execute the following:

> ./build.cmd 

That will print out a list of different commands and the options that they support.

If you want to do a full local build, NuGet packages + tests. Then execute the following

> ./build.cmd all

And then check the `bin/NuGet` folder for  packages. It'll look like this:

image

As part of this PR I've also set up build support on Petabridge's public CI servers (used for OSS builds.) I'm going to run two parallel builds for each DotNetty PR: one for Windows .NET 4.5 and one for Mono 4.0.

If the Mono builds are problematic due to issues with XUnit 2, I can mute those.

I have not set up nightly build support yet - want feedback on the Build proccess and NuGet packaging first.


Aaronontheweb:
If you want to create pre-release NuGet packages, you can just type the following:

```
./build.cmd all nugetprerelease=beta
```

Aaronontheweb:
Looks like I can't rebase the NuGet.exe binary stuff without overwriting the binaries... Issue with it being an unmonitored file in earlier commits. >_<

Aaronontheweb:
And there's a permissions issue with `build.sh` on Mono at the moment - don't have time to research it today, but that'll need to be fixed in order to run the tests on Mono at all.

http://petabridge-ci.cloudapp.net/viewLog.html?buildId=1910&guest=1&buildTypeId=DotNetty_PrMonoBuilds&tab=buildLog#_focus=39

Aaronontheweb:
Made the suggested updated - I'll add a Wiki page for the comments. Waiting on seeing if I fixed all of the issues with the build script on the build server first.

Aaronontheweb:
https://github.com/Azure/DotNetty/wiki/Building-DotNetty

Going to have to press pause on Mono builds for now. Infrastructure is all there but I'm totally puzzled by why TeamCity won't run the bash script. Otherwise, this PR is good to go.

nayato:
looks **great** to me!

Issue #15, opened 2015-08-19
AbstractByteBuffer.WriteBytesAsync(Stream, ...) misuses ArrayOffset


nayato:
ArrayOffset value is being added to WriterIndex at the end while it shouldn't be.

Issue #14, opened 2015-08-19
Enable scheduling work items with 2 state parameters


nayato:
To avoid allocations it is often useful to pass more than one object along with work item: this/self and piece of data to be processed. Such an option is available for immediate execution through IEventExecutor.Execute() overload but there's no corresponding version of .Schedule().

Aaronontheweb:
Ah, ok - this explains my comments on #16 - would you mind adding this explanation in the comments somewhere on the `Schedule` methods?

PR #16, opened 2015-08-19
extra IEventExecutor.Schedule overload, async stream reading fix

nayato:
Adds IEventExecutor.Schedule overload with 2 state parameters per #14
Fix async stream reading to AbstractByteBuffer per #15

### DIFF
--- a/src/DotNetty.Buffers/AbstractByteBuffer.cs
+++ b/src/DotNetty.Buffers/AbstractByteBuffer.cs
@@ -659,15 +659,16 @@ namespace DotNetty.Buffers
         {
             int readTotal = 0;
             int read;
-            int localWriterIndex = this.ArrayOffset + this.WriterIndex;
+            int startWriterIndex = this.WriterIndex;
+            int offset = this.ArrayOffset + startWriterIndex;
             do
             {
-                read = await stream.ReadAsync(this.Array, localWriterIndex + readTotal, length - readTotal, cancellationToken);
+                read = await stream.ReadAsync(this.Array, offset + readTotal, length - readTotal, cancellationToken);
                 readTotal += read;
             }
             while (read > 0 && readTotal < length);
 
-            this.SetWriterIndex(localWriterIndex + readTotal);
+            this.SetWriterIndex(startWriterIndex + readTotal);
             return readTotal;
         }
 
@@ -699,12 +700,18 @@ namespace DotNetty.Buffers
 
         public virtual byte[] ToArray()
         {
-            if (this.HasArray)
+            int readableBytes = this.ReadableBytes;
+            if (readableBytes == 0)
             {
-                return this.Array.Slice(this.ArrayOffset + this.ReaderIndex, this.ReadableBytes);
+                return ByteArrayExtensions.Empty;
             }
 
-            var bytes = new byte[this.ReadableBytes];
+            if (this.HasArray)
+            {
+                return this.Array.Slice(this.ArrayOffset + this.ReaderIndex, readableBytes);
+            }
+
+            var bytes = new byte[readableBytes];
             this.GetBytes(this.ReaderIndex, bytes);
             return bytes;
         }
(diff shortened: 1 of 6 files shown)


Aaronontheweb:
@nayato need to make sure that you rebase on dev and have `build.cmd` locally inside your fork in order for the test suite to run

test/DotNetty.Tests.End2End/End2EndTests.cs
```diff
@@ -60,7 +60,8 @@ public End2EndTests(ITestOutputHelper output)
                 .Option(ChannelOption.TcpNodelay, true)
                 .Handler(new ActionChannelInitializer<ISocketChannel>(ch =>
                 {
-                    ch.Pipeline.AddLast(TlsHandler.Client("dotnetty.com", null, (sender, certificate, chain, errors) => true));
+                    string targetHost = tlsCertificate.GetNameInfo(X509NameType.DnsName, false);
```
Aaronontheweb: could you help me understand what's going on here in this particular segment?

src/DotNetty.Buffers/AbstractByteBuffer.cs
```diff
@@ -659,15 +659,16 @@ public Task<int> WriteBytesAsync(Stream stream, int length, CancellationToken ca
...
             do
             {
-                read = await stream.ReadAsync(this.Array, localWriterIndex + readTotal, length - readTotal, cancellationToken);
+                read = await stream.ReadAsync(this.Array, offset + readTotal, length - readTotal, cancellationToken);
                 readTotal += read;
             }
             while (read > 0 && readTotal < length);
 
-            this.SetWriterIndex(localWriterIndex + readTotal);
+            this.SetWriterIndex(startWriterIndex + readTotal);
```
Aaronontheweb: yep, I think the intent here is much clearer than `localWriterIndex` :+1:

src/DotNetty.Common/Concurrency/IEventExecutor.cs
```diff
@@ -46,5 +46,9 @@ public interface IEventExecutor
         Task ShutdownGracefullyAsync();
 
         Task ShutdownGracefullyAsync(TimeSpan quietPeriod, TimeSpan timeout);
+
+        void Schedule(Action<object, object> action, object state1, object state2, TimeSpan delay);
+
+        void Schedule(Action<object, object> action, object state1, object state2, TimeSpan delay, CancellationToken cancellationToken);
```
Aaronontheweb: What's the purpose of having two different state objects in the `Schedule` signature here? Does one have a distinct difference than the other? Or is it totally arbitrary and specific to the `Action` ?

Not challenging the current design, just want to know what's the distinction here :p

test/DotNetty.Tests.End2End/End2EndTests.cs
nayato: SslStream needs targetHost which is basically a DNS name so we're pulling that out of cert. It's test cert so it was OK to just hard-code it but is in general better to get info from the cert itself.

src/DotNetty.Common/Concurrency/IEventExecutor.cs
Aaronontheweb: Got it - just read your explanation on #14 

IMHO, add this as a comment:

> To avoid allocations it is often useful to pass more than one object along with work item: this/self and piece of data to be processed. Such an option is available for immediate execution through IEventExecutor.Execute() overload but there's no corresponding version of .Schedule().

test/DotNetty.Tests.End2End/End2EndTests.cs
Aaronontheweb: Got it, makes sense

nayato:
@Aaronontheweb please, take a look. Added xml doc on IEventExecutor. I'll squash commits before merging. Right now keeping them separate helps with review.

Issue #23, opened 2015-09-17
Nuget packages produced by build carry no assemblies


nayato:
Build using `build.cmd all`: nuget packages are created but have no `lib` folder in them.

PR #24, opened 2015-09-18
Fixes nuget package creation build target (#23)

nayato:
Generated nuget packages now contain proper assemblies.
+ changed author to Microsoft Azure
+ dereferencing SslStream-provided state when is not needed anymore in TlsHandler.MediationStream.


mastermanu:
LGTM

PR #25, opened 2015-09-18
Add a new Buffer Pool feature

### LINKED ISSUE #9: Better pooled byte buffer allocator
Port pooled byte buffer allocator from netty. .NET does not have a concept of JVM's ByteBuffer so it's OK to collapse impl to byte array version only (Heap...).

zhengchun:
fixes  #9 
New Buffer Pooled feature,but not implement Thread Local cache yet.

### DIFF
--- a/src/DotNetty.Buffers/AbstractReferenceCountedByteBuffer.cs
+++ b/src/DotNetty.Buffers/AbstractReferenceCountedByteBuffer.cs
@@ -10,7 +10,7 @@ namespace DotNetty.Buffers
     public abstract class AbstractReferenceCountedByteBuffer : AbstractByteBuffer
     {
 #pragma warning disable 420
-        volatile int _refCounted = 1;
+        volatile int referenceCount = 1;
 
         protected AbstractReferenceCountedByteBuffer(int maxCapacity)
             : base(maxCapacity)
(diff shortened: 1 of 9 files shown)


nayato:
Hi! Thanks for taking on this! What version is this based on? We're porting from netty's master and it's quite a bit different from your PR.

zhengchun:
based on Netty 5.0,i know some code is quite different with Netty,because i removed heap allocate(i don't consider need a unmanaged memory allocate in .NET) and changed some function name,but the main code and functional like netty.

nayato:
@zhengchun, I've made a first round of review. It's quite a bit harder to review with all the deviations from original source code, and it's not a trivial piece of code. Also, I'll reiterate, going forward we'd really want code to be as close to the original as possible to benefit from co-development of netty (which is super-active all the time). It would be great if you'd be able to do a little extra to add the missing pieces (`PoolThreadCache`, `PoolChunkList` and `qXXX` approach in `PoolArena`). Otherwise, having a list of significant changes you had to make to simplify the code in this PR would be indispensable if someone else would have to continue this effort.

zhengchun:
@nayato,thanks for code review,in the past few months,i was developing a asynchronous TCP framework that inspired from Netty and take it as my open source,so simplify design and better for .Net developer.
A weeks ago,i found this project and very similarly with my project ,so i decide pull request code from my project,it's simplified version of Buffer pool,I think these code may be can help you to this project.

PR #33, opened 2015-10-06
Echo client-server example

Horusiath:
This is example of DotNetty client-server application. Server itself can be called using telnet. IMO missing parts are logging handler (which I havent found inside DotNetty implementation) and SSL support which is handled differently than using corellated netty examples.

### DIFF
--- a/DotNetty.sln
+++ b/DotNetty.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.23107.0
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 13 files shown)


nayato:
@Horusiath, on SSL - have you seen TlsHandler? It's meant as SslHandler replacement based on SslStream. It's implemented differently but was meant to behave in the same way.

Horusiath:
I've update example with SSL support.

nayato:
Thanks! will surely review this tomorrow (PST)

Horusiath:
I've updated the example with @nayato suggestions - only thing left is echo client printing out message echo'ed from server.

nayato:
@Horusiath, LoggingHandler is a must in many situations especially with non-trivial pipeline composition so it will definitely be ported soon (#38). It actually provides nice looking bytebuf output + additional log for all the events. I'd say, that's how one should learn netty / DotNetty - by observing output from logging handler. Once that's in, we should use it in examples.

nayato:
pls fix the commit message. I think we should move to netty style for this: http://netty.io/wiki/writing-a-commit-message.html

nayato:
also, have you signed our CLA? https://cla.azure.com/

azurecla:
__@Horusiath__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

Issue #39, opened 2015-10-14
port InternalLogger/Factory


nayato:
`EventSource`-first approach does not seem to scale (in development terms) well and is not friendly for 3rd party loggers.
With InternalLogger back, it should be easier to port other parts of netty. + It will become easier to configure logging through use of a single `EventSource` as a default logging "channel".

PR #41, opened 2015-10-21
Port of InternalLogger (#39)

nayato:
Motivation:
enable netty's internal logging experience to ease porting efforts and streamline logging configuration

Modifications:
- ported over InternalLogger and InternalLoggerFactory
- implemented InternalLogger using EventSource
- switched existing internal logging to this (increasing parity with netty)
- added logging configuration in Echo example
- added log to xunit output to end-to-end tests

*** extras
- fixed Echo client to send back received message (not 0 bytes)
- MqttDecoder no longer closes the Channel by itself instead changing internal state to invalid and forwarding an exception.

Result:
internal components can use familiar experience to perform logging, e.g. `static readonly IInternalLogger Logger = InternalLoggerFactory.GetInstance<AbstractChannel>();`


affandar:
looks great. thanks

PR #44, opened 2015-11-03
fixed casing issue with Mono build script

Aaronontheweb:
This should allow the Mono builds to run again, combined with some other fixes I made on TeamCity

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,19 +9,19 @@ cd `dirname ${SCRIPT_PATH}` > /dev/null
 SCRIPT_PATH=`pwd`;
 popd  > /dev/null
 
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Looks like I need to update the version of Mono on the build server agents. Working on that now.

Aaronontheweb:
Disabling Mono builds - Mono 4.0+ just isn't in a state where it can run anything substantial right now using our buildchain.

Aaronontheweb:
Whoops, meant to reopen it - we should still pull in these changes but there are bigger issues at work stopping Mono builds from running properly beyond what's in this script.

PR #48, opened 2015-11-06
Porting resource leak detection

nayato:
Motivation:
Leak of IReferenceCounted resources is a very common issue while
developing with DotNetty (both codecs and higher level applications).
Having leak detection helps identifying and resolving this kind of issues.

Modifications:
- Added ResourceLeakDetector
- Added leak-aware Byte Buffer implementations

extras
- port of missing Byte Buffer methods
- fixing minor issues with previous PRs

Result:
Simple leak detection is enabled by default on PooledByteBufferAllocator,
leaks are reported through internal logger, advanced leak detection is
available via LeakResourceDetector


src/DotNetty.Common/Deque.cs
```diff
@@ -0,0 +1,875 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+
+namespace Nito
```
Aaronontheweb: What's up with the namespace here?

src/DotNetty.Common/IResourceLeak.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace DotNetty.Common
+{
+    public interface IResourceLeak
+    {
+        /// <summary>
+        ///     Records the caller's current stack trace so that the <see cref="ResourceLeakDetector" /> can tell where the
+        ///     leaked
```
Aaronontheweb: Whitespace

src/DotNetty.Common/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Nito.Deque" version="1.0.0" targetFramework="net45" />
```
Aaronontheweb: So are we taking a dependency on Nito?

nayato: Do you have any better Deque impl in mind?

src/DotNetty.Common/Deque.cs
nayato: File comes from the nuget.

src/DotNetty.Common/IResourceLeak.cs
nayato: ?

Aaronontheweb: You just have an extra linebreak here that breaks the comment onto two lines - it just looks odd

src/DotNetty.Common/packages.config
Aaronontheweb: So inline with the previous question then, this NuGet package just adds a source file and no actual binary?

nayato: correct

Aaronontheweb: Remove this line from the packages.config file or adjust the FAKE script so it doesn't accidentally include Nito as a dependency in our own NuGet package then (since FAKE builds the NuSpec assuming that everything in packages.config for this project is a dependency)

PR #56, opened 2015-11-25
Fix #55

### LINKED ISSUE #55: DotNetty.Buffers IndexOutOfBoundsException on ChannelGroup.WriteAndFlushAsync(Unpooled.Empty)
Haven't determined what the underlying cause of the error is yet, but the following block of code throws an `IndexOutOfBoundsException` inside the `Task` returned on `ChannelGroup.WriteAndFlushAsync`

```csharp
public override async Task<bool> Shutdown()
{
    Func<Task, bool> always = task => task.IsCompleted & !task.IsCanceled & !task.IsFaulted;

    var writeStatus = ChannelGroup.WriteAndFlushAsync(Unpooled.Empty).ContinueWith(always); //throws inside Task
    var disconnectStatus = ChannelGroup.DisconnectAsync().ContinueWith(always);
    var closeStatus = ChannelGroup.CloseAsync().ContinueWith(always);
    //_eventListener.Dispose();
    return await Task.WhenAll(writeStatus, disconnectStatus, closeStatus).ContinueWith(tr =>
    {
        if (tr.IsFaulted || tr.IsCanceled) return false;
        return tr.Result.All(x => x);
    });
}
```

nayato:
Motivation:
revisit EmptyByteBuffer port to ensure complete functionality

Modifications:
- ported trivial implementation of ToArray, Duplicate and other methods
- minor cleanup

Result:
EmptyByteBuffer does not throw exceptions when operation can be completed in a trivial way (0 byte length is enough to complete the operation).

### DIFF
--- a/src/DotNetty.Transport/Channels/Groups/DefaultChannelGroup.cs
+++ b/src/DotNetty.Transport/Channels/Groups/DefaultChannelGroup.cs
@@ -310,18 +310,19 @@ namespace DotNetty.Transport.Channels.Groups
 
         static object SafeDuplicate(object message)
         {
-            if (message is IByteBuffer)
+            var buffer = message as IByteBuffer;
+            if (buffer != null)
             {
-                return ((IByteBuffer)message).Duplicate().Retain();
+                return buffer.Duplicate().Retain();
             }
-            else if (message is IByteBufferHolder)
+
+            var byteBufferHolder = message as IByteBufferHolder;
+            if (byteBufferHolder != null)
             {
-                return ((IByteBufferHolder)message).Duplicate().Retain();
+                return byteBufferHolder.Duplicate().Retain();
             }
-            else
-            {
-                return ReferenceCountUtil.Retain(message);
-            }
+
+            return ReferenceCountUtil.Retain(message);
         }
 
         public override bool Equals(object obj)
@@ -355,13 +356,14 @@ namespace DotNetty.Transport.Channels.Groups
 
         public bool Remove(IChannelId channelId)
         {
-            IChannel ch = null;
+            IChannel ch;
 
             if (this.serverChannels.TryRemove(channelId, out ch))
             {
                 return true;
             }
-            else if (this.nonServerChannels.TryRemove(channelId, out ch))
+
+            if (this.nonServerChannels.TryRemove(channelId, out ch))
             {
                 return true;
             }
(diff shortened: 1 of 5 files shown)


nayato:
@Aaronontheweb PTAL

src/DotNetty.Transport/Channels/Groups/DefaultChannelGroup.cs
```diff
@@ -310,18 +310,19 @@ public Task NewCloseFuture(IChannelMatcher matcher)
 
         static object SafeDuplicate(object message)
         {
-            if (message is IByteBuffer)
+            var buffer = message as IByteBuffer;
+            if (buffer != null)
             {
-                return ((IByteBuffer)message).Duplicate().Retain();
+                return buffer.Duplicate().Retain();
```
Aaronontheweb: IMHO, we should really have a test case to verify that this actually fixes #55

src/DotNetty.Transport/Channels/Groups/DefaultChannelGroupCompletionSource.cs
```diff
@@ -34,7 +34,7 @@ public DefaultChannelGroupCompletionSource(IChannelGroup group, Dictionary<IChan
                 this.futures.Add(pair.Key, pair.Value);
                 pair.Value.ContinueWith(x =>
                 {
-                    bool success = x.IsCompleted && !x.IsFaulted && !x.IsCanceled;
+                    bool success = x.Status == TaskStatus.RanToCompletion;
```
Aaronontheweb: just for my own benefit, this statement and the previous statement are equivalent? "Completed with no errors or cancellation?"

src/DotNetty.Transport/Channels/Groups/DefaultChannelGroup.cs
nayato: this change specifically does nothing, it's minor cleanup. Actual change to fix out of range exceptions are in `EmptyByteBuffer`

src/DotNetty.Transport/Channels/Groups/DefaultChannelGroupCompletionSource.cs
nayato: correct, only difference is that we're doing a single volatile read with the updated version.

mtuchkov:
It must be OK. In offline discussion with @nayato we agreed that we need a general-purpose test that would cover greater number of cases such as byte buffers of different sizes, including corner cases like 0 bytes, different presets, etc. That can be a separate work item.

mtuchkov:
Please bump the version and update the nuspec as well.

PR #62, opened 2016-01-24
Implemented ToString methods in ByteBuffer for user convenience.

hide1202:
I implemented ToString methods in ByteBuffer for user convenience. And modified echo examples.
These methods were implemented in Netty(Java).

# Java version
```java
// In ByteBuf.java
public abstract String toString(Charset charset);
public abstract String toString(int index, int length, Charset charset);
```

# C# version
```C#
// In IByteBuffer.cs
string ToString(Encoding encoding);
string ToString(int index, int length, Encoding encoding);
```

# Usages in EchoClientHandler.cs
## Before
```C#
this.buffer.Initialize();
byteBuffer.Duplicate().ReadBytes(this.buffer, 0, byteBuffer.ReadableBytes);
string msg = Encoding.UTF8.GetString(this.buffer);
Console.WriteLine("Received from server: " + msg);
```

## After
```C#
Console.WriteLine("Received from server: " + byteBuffer.ToString(Encoding.Default));
```

### DIFF
--- a/examples/Echo.Client/EchoClientHandler.cs
+++ b/examples/Echo.Client/EchoClientHandler.cs
@@ -31,11 +31,8 @@ namespace Echo.Client
         {
             var byteBuffer = message as IByteBuffer;
             if (byteBuffer != null)
-            {
-                this.buffer.Initialize();
-                byteBuffer.Duplicate().ReadBytes(this.buffer, 0, byteBuffer.ReadableBytes);
-                string msg = Encoding.UTF8.GetString(this.buffer);
-                Console.WriteLine("Received from server: " + msg);
+            {                
+                Console.WriteLine("Received from server: " + byteBuffer.ToString(Encoding.UTF8));
             }
             context.WriteAsync(message);
         }
--- a/examples/Echo.Server/EchoServerHandler.cs
+++ b/examples/Echo.Server/EchoServerHandler.cs
@@ -3,12 +3,19 @@
 namespace Echo.Server
 {
     using System;
+    using System.Text;
+    using DotNetty.Buffers;
     using DotNetty.Transport.Channels;
 
     public class EchoServerHandler : ChannelHandlerAdapter
     {
         public override void ChannelRead(IChannelHandlerContext context, object message)
         {
+            IByteBuffer buffer = message as IByteBuffer;
+            if (buffer != null)
+            {
+                Console.WriteLine("Received from client: " + buffer.ToString(Encoding.UTF8));
+            }            
             context.WriteAsync(message);
         }
 
--- a/src/DotNetty.Buffers/ByteBufferUtil.cs
+++ b/src/DotNetty.Buffers/ByteBufferUtil.cs
@@ -569,5 +569,32 @@ namespace DotNetty.Buffers
                 dump.Append('|');
             }
         }
+
+        public static string DecodeString(IByteBuffer src, int readerIndex, int len, Encoding encoding)
+        {
+            if (len == 0)
+            {
+                return string.Empty;
+            }
+
+            if (src.HasArray)
+            {
+                return encoding.GetString(src.Array, readerIndex, len);
+            }
+            else
+            {
+                IByteBuffer buffer = src.Allocator.Buffer(len);
+                try
+                {
+                    buffer.WriteBytes(src, readerIndex, len);
+                    return encoding.GetString(buffer.Array, 0, len);
+                }
+                finally
+                {
+                    // Release the temporary buffer again.
+                    buffer.Release();
+                }
+            }
+        }
     }
 }
--- a/src/DotNetty.Buffers/EmptyByteBuffer.cs
+++ b/src/DotNetty.Buffers/EmptyByteBuffer.cs
@@ -6,6 +6,7 @@ namespace DotNetty.Buffers
     using System;
     using System.Diagnostics.Contracts;
     using System.IO;
+    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
     using DotNetty.Common;
@@ -643,6 +644,17 @@ namespace DotNetty.Buffers
                 throw new IndexOutOfRangeException();
             }
             return this;
+        }
+
+        public string ToString(Encoding encoding)
+        {
+            return string.Empty;
+        }
+
+        public string ToString(int index, int length, Encoding encoding)
+        {
+            CheckIndex(index, length);
+            return this.ToString(encoding);
         }
     }
 }
--- a/src/DotNetty.Buffers/AbstractByteBuffer.cs
+++ b/src/DotNetty.Buffers/AbstractByteBuffer.cs
@@ -6,6 +6,7 @@ namespace DotNetty.Buffers
     using System;
     using System.Diagnostics.Contracts;
     using System.IO;
+    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
     using DotNetty.Common;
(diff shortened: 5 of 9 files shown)


azurecla:
Hi __@hide1202__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla.azure.com.
    </span>

TTYL, AZPRBOT;

azurecla:
__@hide1202__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

src/DotNetty.Buffers/EmptyByteBuffer.cs
```diff
@@ -644,5 +645,15 @@ IByteBuffer CheckLength(int length)
...
         }
+
+        public string ToString(Encoding encoding)
+        {
+            return string.Empty;
+        }
+
+        public string ToString(int index, int length, Encoding encoding)
+        {
+            return string.Empty;
```
nayato: Follow netty in this:
```
        checkIndex(index, length);
        return toString(charset);
```

src/DotNetty.Buffers/ByteBufferUtil.cs
```diff
@@ -569,5 +569,12 @@ static void AppendHexDumpRowPrefix(StringBuilder dump, int row, int rowStartInde
                 dump.Append('|');
             }
         }
+
+        public static string DecodeString(IByteBuffer src, int readerIndex, int len, Encoding encoding)
+        {
+            if (len == 0) return string.Empty;
+            
+            return encoding.GetString(src.Array.Slice(len));
```
nayato: buffer may not have a backing array (i.e. `buf.HasArray == false`). CompositeByteBuffer is a good example.

src/DotNetty.Buffers/EmptyByteBuffer.cs
hide1202: OK, I will modify this.

src/DotNetty.Buffers/ByteBufferUtil.cs
nayato: probably the easiest way to address this is to check if `buf.HasArray == false` and in this case, alloc new buffer from allocator and write necessary range of bytes there. It still implies that allocator should return buffer with `buf.HasArray == true` but I think it's OK to assume that. Now, if `buf.HasArray == true` you can use overload of GetString that decodes string from range of byte array: https://msdn.microsoft.com/en-us/library/05cts4c3(v=vs.110).aspx

hide1202: @nayato I think that this case is not necessary to allocate new ByteBuffer. If ```buf.HasArray==false```, just return string.Empty.
So i understand like following code.
How do you think about this code??
```C#
if (src.HasArray == false || len == 0)
{
	return string.Empty;
}
return encoding.GetString(src.Array, readerIndex, len);
```

nayato: `HasArray` is actually an indicator, whether the byte buffer's content is available through `Array` property directly. So for `CompositeByteBuffer` with 2+ components it will be `false`. so really in case of `buf.HasArray == false` you'd need to read it out into array-backed buffer.

nayato:
Also, please add override in `AdvancedLeakAwareByteBuf` same as in netty.

examples/Echo.Client/EchoClientHandler.cs
```diff
@@ -31,11 +31,8 @@ public override void ChannelRead(IChannelHandlerContext context, object message)
         {
             var byteBuffer = message as IByteBuffer;
             if (byteBuffer != null)
-            {
-                this.buffer.Initialize();
-                byteBuffer.Duplicate().ReadBytes(this.buffer, 0, byteBuffer.ReadableBytes);
-                string msg = Encoding.UTF8.GetString(this.buffer);
-                Console.WriteLine("Received from server: " + msg);
+            {                
+                Console.WriteLine("Received from server: " + byteBuffer.ToString(Encoding.Default));
```
nayato: `.ToString(Encoding.UTF8)` I think would be best suited for example here.

examples/Echo.Server/EchoServerHandler.cs
```diff
@@ -3,12 +3,19 @@
...
     using DotNetty.Transport.Channels;
 
     public class EchoServerHandler : ChannelHandlerAdapter
     {
         public override void ChannelRead(IChannelHandlerContext context, object message)
         {
+            IByteBuffer buffer = message as IByteBuffer;
+            if (buffer != null)
+            {
+                Console.WriteLine("Received from client: " + buffer.ToString(Encoding.Default));
```
nayato: `Encoding.UTF8`?

examples/Echo.Client/EchoClientHandler.cs
hide1202: OK~!!

examples/Echo.Server/EchoServerHandler.cs
hide1202: OK~!!

hide1202: I will add ToString overrides in AdvancedLeakAwareByteBuf same as in netty.

src/DotNetty.Buffers/ByteBufferUtil.cs
hide1202: @nayato Sorry, i didn't understand 'CompositeByteBuffer'. But ToString of AbstractByteBuf is expansive method. I think that ToString CompositeByteBuffer is special case, so this method will be overrided. But Netty(Java) don't implemented ToString of CompositeByteBuffer. So i think that this implementation is TBD.
And according to reference Netty(Java), I think that HasArray check in ToString should be removed.

nayato: In netty there's a lot of plumbing around direct vs heap buffers which does not quite (yet) applies in DotNetty.
Check this line: https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L541
This is quite close in semantics to `HasArray` in DotNetty. It's one of those methods where it's impossible to follow netty 1:1 just because underlying framework is different :)

nayato: To clarify, with distinction based on `HasArray` `DecodeString` would still match the behavior of netty in generic way that is also most efficient for the common case (byte array backed buffer). There's no need to implement special ToString on CompositeByteBuffer at this point.

hide1202: @nayato Oops! sorry :(
I re-write code again. Please, check this code.
```C#
public static string DecodeString(IByteBuffer src, int readerIndex, int len, Encoding encoding)
{
	if (len == 0) return string.Empty;
	
	if (src.HasArray)
	{
		return encoding.GetString(src.Array, readerIndex, len);
	}
	else
	{
		IByteBuffer buffer = src.Allocator.Buffer(len);
		try
		{
			buffer.WriteBytes(src, readerIndex, len);
			return encoding.GetString(buffer.Array, 0, len);
		}
		finally
		{
			// Release the temporary buffer again.
			buffer.Release();
		}
	}
}
```

nayato: looks good. Just format the `if (len...` line with braces per coding style.

hide1202: @nayato  Oops! my mistake~ I will correct and push.

hide1202:
@nayato  Fixed codes. Please check commits.

nayato:
Looks good! now, few little things remain:
- collapse commits into one (http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)
- give it a proper commit message (http://netty.io/wiki/writing-a-commit-message.html).

and I'll happily merge. Thanks for taking on this!

hide1202:
@nayato  thank you~! I rebase commits into one, and modify commit message. Please check~ :)

nayato:
merged. Thanks again!

PR #67, opened 2016-02-10
Fixes premature completion of async operations (#66)

nayato:
Motivation:
Due to misuse of IEventExecutor.SubmitAsync in
DefaultChannelHandlerInvoker and Bootstrap, async operations return future
that is completed after first sync part of operation is through.

Modifications:
DefaultChannelHandlerInvoker and Bootstrap are modified to complete
returned future only after the whole async operation is completed.

Result:
chaining async operations does not result in intermittent failures due to
premature access to channel, errors potentially appearing in later stages
of async operations are also properly reported through failure of the
future.

### DIFF
--- a/src/DotNetty.Common/Concurrency/TaskCompletionSource.cs
+++ b/src/DotNetty.Common/Concurrency/TaskCompletionSource.cs
@@ -34,6 +34,11 @@ namespace DotNetty.Common.Concurrency
             return true;
         }
 
+        public override string ToString()
+        {
+            return "TaskCompletionSource[status: " + this.Task.Status.ToString() + "]";
+        }
+
         static TaskCompletionSource CreateVoidTcs()
         {
             var tcs = new TaskCompletionSource();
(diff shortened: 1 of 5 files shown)


src/DotNetty.Common/Concurrency/TaskCompletionSource.cs
```diff
@@ -34,6 +34,11 @@ public bool setUncancellable()
             return true;
         }
 
+        public override string ToString()
+        {
+            return "TaskCompletionSource[status: " + this.Task.Status.ToString() + "]";
```
manusr: + .ToString() is redundant, right? It should auto-convert based on inferring the '+' op

manusr:
lgtm

src/DotNetty.Common/Concurrency/TaskCompletionSource.cs
nayato: it's `Concat(string..)` vs `Concat(object..)`. I prefer to explicitly call `ToString` to avoid hidden boxing. In this case though, it'll get boxed anyway (`Enum.ToString()` issue) but this is for debugging / unhappy path logging only anyway.

PR #69, opened 2016-02-10
Enables setting options specific to DefaultSocketChannelConfiguration…

### LINKED ISSUE #68: DefaultSocketChannelConfiguration.SetOption is broken
`DefaultSocketChannelConfiguration` is missing `SetOption` overload resulting in failure to accept Socket-specific options through generic `SetOption` way.

nayato:
… (fix #68)

Motivation:
Fix a bug that breaks setting options specific to socket-based channel implementations

Modifications:
Added SetOption overload accepting socket-specific options to DefaultSocketChannelConfiguration

Result:
It is possible to modify send/receive buffer size and other settings through SetOption on Bootstrap.


mtuchkov:
Legitimate

Issue #71, opened 2016-02-17
Pooled buffer's AdjustCapacity breaks on fall-through to unpooled implementation


nayato:
Buffer's length is not updated if buffer's growth results in a switch to unpooled underlying array. That leads to failures in consecutive operations involving access to buffer's area higher than original length.

PR #72, opened 2016-02-17
Fixes pooled buffer's AdjustCapacity to properly set the length

nayato:
Motivation:
Fix a bug #71.

Modifications:
Set length property of a PooledByteBuffer in case of a switch of underlying array to unpooled.

Results:
PooledByteBufferAllocator can work with buffers bigger than pooled buffer's original capacity.


manusr:
LGTM

PR #80, opened 2016-04-04
Adds IEventExecutor.Schedule, proper cancellation of scheduled tasks

nayato:
Motivation:
Allow for less GC when scheduling if Task-based API is not required. Allow for proper cleanup when canceling scheduled task so that associated resources can be cleaned up early.

Modifications:
- Moved out scheduled task implementations from AbstractScheduledEventExecutor
- Introduced IScheduledTask and aligned IScheduledRunnable
- Implemented removal of scheduled task upon its cancellation

Result:
Scheduled tasks that normally never fire (e.g. timeout handling) do not cause resources to be held up unnecessarily in memory for longer period of time if not necessary.

### DIFF
--- a/src/DotNetty.Common/Concurrency/ScheduledAsyncTask.cs
+++ b/src/DotNetty.Common/Concurrency/ScheduledAsyncTask.cs
@@ -1,0 +1,33 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace DotNetty.Common.Concurrency
+{
+    using System.Threading;
+
+    abstract class ScheduledAsyncTask : ScheduledTask
+    {
+        readonly CancellationToken cancellationToken;
+        CancellationTokenRegistration cancellationTokenRegistration;
+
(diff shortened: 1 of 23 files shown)


src/DotNetty.Common/Concurrency/ScheduledAsyncTask.cs
```diff
@@ -0,0 +1,41 @@
...
+                    task.Executor.Execute(t => ((ScheduledAsyncTask)t).Cancel(), task);
+                }
+            }, this);
+        }
+
+        public override void Run()
+        {
+            if (this.cancellationToken.IsCancellationRequested)
+            {
+                this.Promise.TrySetCanceled();
```
mtuchkov: return from here

src/DotNetty.Common/Concurrency/ScheduledTask.cs
```diff
@@ -0,0 +1,113 @@
...
+                {
+                    // todo: check for fatal
+                    this.Promise.TrySetException(ex);
+                }
+            }
+        }
+
+        protected abstract void Execute();
+
+        bool TrySetUncancelable()
```
mtuchkov: Please re-use this method above

mtuchkov:
Looks good

PR #81, opened 2016-04-06
STEE: smarter idle waiting for scheduled tasks

nayato:
Motivation:
use of breakoutInterval setting on SingleThreadEventExecutor while executor is idling is confusing to users and inefficient due to fixed interval wake ups. We can do better by peeking next scheduled task and use its deadline for wait timeout.
While breakoutInterval is still useful while executor is busy, it can be reduced now to much lower number without impacting idle performance.

Modifications:
- When going to wait due to empty immediate queue, consult with scheduled task queue to figure out timeout value (next task's deadline - now).
- default breakoutInterval changed to 100 ms for higher resolution for scheduled tasks while busy.

Result:
Better precision for scheduled tasks at no performance cost while idling.


mtuchkov:
Sounds good

PR #82, opened 2016-04-08
Better signed build automation

nayato:
Motivation
Since signed assemblies became part of the release they slowed down release cycle due to extended effort necessary to prepare nuget packages with signed assemblies.

Modifications
- introduced "Signed" solution configuration
- optional Test Friends (turned off in Signed)
- re-signing support in build script

Result
New releases can be produced quicker with less effort.


nayato:
@Aaronontheweb, you'd be the best person to check build.sfx changes.

Aaronontheweb:
@nayato I'll take a look

Aaronontheweb:
Looks good to me - the .SLN configuration takes care of the nasty stuff, but the`NugetSigned` steps and its sub-steps in FAKE are correct.

PR #83, opened 2016-04-15
Hotfix: poll task from the queue.

mtuchkov:
Fixed polling task from the queue.

Motivation:

Polling task from queue is calculating timeout incorrectly.

Modifications:

Timeout calculation has been fixed

Result:

Event executer works as it should now

### DIFF
--- a/src/DotNetty.Common/PreciseTimeSpan.cs
+++ b/src/DotNetty.Common/PreciseTimeSpan.cs
@@ -15,8 +15,14 @@ namespace DotNetty.Common
         readonly long ticks;
 
         PreciseTimeSpan(long ticks)
+            : this()
         {
             this.ticks = ticks;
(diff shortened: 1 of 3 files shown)


src/DotNetty.Common/PreciseTimeSpan.cs
```diff
@@ -15,10 +15,16 @@ public struct PreciseTimeSpan : IComparable<PreciseTimeSpan>, IEquatable<Precise
         readonly long ticks;
 
         PreciseTimeSpan(long ticks)
+            : this()
         {
             this.ticks = ticks;
         }
 
+        public long Ticks
```
nayato: let's limit its use to API only and use `this.ticks` internally

nayato:
looking good

PR #102, opened 2016-05-05
switching to C# 6.0

nayato:


mtuchkov:
Go go go! Resharper should not let us down. >.<

PR #108, opened 2016-05-12
Enables multi-buffer send in TcpSocketChannel. Remove constraints on …

nayato:
…IByteBuffer for writing.

Motivation:
In order to avoid buffer copies and yet perform IO efficiently,
it is essential to support sending multiple byte sequences in one call.
Also there should be no additional constraints on `IByteBuffer` implementations
in order to be accepted by `TcpSocketChannel` for sending.

Modifications:
- `IoBufferCount`, `GetIoBuffer()`, `GetIoBuffers()` are ported to `IByteBuffer` and implementations.
- `ChannelOutboundBuffer` is extended to leverage `*IoBuffer*` members in order to prepare buffer batches for sending.
- `TcpSocketChannel` and its ancestors were extended to support sending (sync and async) a list of array segments.
- Extra: fixes `WriterIndex` positioning in `ByteBufferUtil.EncodeString(..)`.

Result:
A benchmark of single client channel writing to a single server channel runs 3 times faster at 150K op/sec (vs 53K op/sec before) with no increase in CPU consumption.
`CompositeByteBuffer` can be used directly in `IChannel.WriteAsync`.

### DIFF
--- a/src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
+++ b/src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
@@ -4,6 +4,7 @@
 namespace DotNetty.Transport.Channels
 {
     using System;
+    using System.Collections.Generic;
     using System.Diagnostics.Contracts;
     using System.Threading;
     using DotNetty.Buffers;
@@ -18,7 +19,7 @@ namespace DotNetty.Transport.Channels
 
         static readonly IInternalLogger Logger = InternalLoggerFactory.GetInstance<ChannelOutboundBuffer>();
 
-        //static readonly ThreadLocal<IByteBuffer[]> NIO_BUFFERS = new ThreadLocal<IByteBuffer[]>(() => new IByteBuffer[1024]);
+        static readonly ThreadLocalByteBufferList NioBuffers = new ThreadLocalByteBufferList();
 
         readonly IChannel channel;
 
@@ -34,7 +35,6 @@ namespace DotNetty.Transport.Channels
         int flushed;
 
         //int nioBufferCount;
-        //long nioBufferSize;
 
         bool inFail;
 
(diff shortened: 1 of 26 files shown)


nayato:
@mastermanu, @myagley, @mtuchkov please take a look.

src/DotNetty.Transport/Channels/Sockets/TcpSocketChannel.cs
```diff
@@ -1,9 +1,10 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Licensed under the MIT license. See LICENSE file input the project root for full license information.
```
Aaronontheweb: Search and replace error? :p

src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
```diff
@@ -344,11 +306,104 @@ public void RemoveBytes(long writtenBytes)
...
+                            break;
+                        }
+                        nioBufferSize += readableBytes;
+                        int count = entry.Count;
+                        if (count == -1)
+                        {
+                            //noinspection ConstantValueVariableUse
+                            entry.Count = count = buf.IoBufferCount;
+                        }
+                        if (count == 1)
```
myagley: In the netty version of this function, there appears to be some code to expand the array of buffers if the needed space is greater then the current length of the buffers.

https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java#L410

nayato: I'm using list which does that internally. And I need to use list because Socket.Send accepts IList<ArraySegment<byte>>> meaning that there's no way to supply different length along with array.

myagley: Ok. I overlooked the use of list.

myagley:
Looks good to me.

PR #113, opened 2016-05-17
added DotNetty stress spec

Aaronontheweb:
This spec will continuously add new client connections (firing roughly 255 msg / s each) to the same server until the time to add 10 connections exceeds the pre-defined time threshold, which currently stands at 15 seconds.

Goal is to determine how many concurrent connections can be added until we've saturated the CPU. Bear in mind this spec is running both the client and the server load on the same machine, so in real life the connection threshold for a single server would be much higher.

### DIFF
--- a/test/DotNetty.Transport.Tests.Performance/Sockets/TcpServerSocketChannelHorizontalScaleStressSpec.cs
+++ b/test/DotNetty.Transport.Tests.Performance/Sockets/TcpServerSocketChannelHorizontalScaleStressSpec.cs
@@ -1,0 +1,196 @@
+﻿namespace DotNetty.Transport.Tests.Performance.Sockets
+{
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Net;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using DotNetty.Codecs;
+    using DotNetty.Transport.Bootstrapping;
+    using DotNetty.Transport.Channels;
+    using DotNetty.Transport.Channels.Sockets;
+    using DotNetty.Transport.Tests.Performance.Utilities;
+    using NBench;
+
+    public class TcpServerSocketChannelHorizontalScaleSpec
+    {
+        const string ClientConnectCounterName = "connected clients";
+
+        const string InboundThroughputCounterName = "inbound ops";
+
+        const string OutboundThroughputCounterName = "outbound ops";
+
+        const string ErrorCounterName = "exceptions caught";
+
+        public const int IterationCount = 1; // these are LONG-running benchmarks. stick with a lower iteration count
+        static readonly IPEndPoint TEST_ADDRESS = new IPEndPoint(IPAddress.IPv6Loopback, 0);
+
+        // Sleep main thread, then start a new client every 30 ms
+        static readonly TimeSpan SleepInterval = TimeSpan.FromMilliseconds(300);
+
+        /// <summary>
+        ///     If it takes longer than <see cref="SaturationThreshold" /> to establish 10 connections, we're saturated.
+        ///     End the stress test.
+        /// </summary>
+        static readonly TimeSpan SaturationThreshold = TimeSpan.FromSeconds(15);
+
+        protected readonly ManualResetEventSlim ResetEvent = new ManualResetEventSlim(false);
+        ConcurrentBag<IChannel> clientChannels;
+        Counter clientConnectedCounter;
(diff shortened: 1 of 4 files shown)


azurecla:
Hi __@Aaronontheweb__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <span>We will now validate the agreement and then real humans will evaluate your PR.</span>

TTYL, AZPRBOT;

test/DotNetty.Transport.Tests.Performance/Sockets/TcpServerSocketChannelHorizontalScaleStressSpec.cs
```diff
@@ -0,0 +1,196 @@
...
+        protected virtual IChannelHandler GetEncoder() => new LengthFieldPrepender(4, false);
+
+        protected virtual IChannelHandler GetDecoder() => new LengthFieldBasedFrameDecoder(int.MaxValue, 0, 4, 0, 4);
+
+        static readonly ThreadLocal<Random> ThreadLocalRandom = new ThreadLocal<Random>(() => new Random());
+
+        [PerfSetup]
+        public void SetUp(BenchmarkContext context)
+        {
+            this.ClientGroup = new MultithreadEventLoopGroup(Environment.ProcessorCount / 2);
```
nayato: have you tried making it ProcessorCount ?

test/DotNetty.Transport.Tests.Performance/Sockets/TcpServerSocketChannelHorizontalScaleStressSpec.cs
```diff
@@ -0,0 +1,196 @@
...
+                .ChildOption(ChannelOption.TcpNodelay, true)
+                .ChildHandler(new ActionChannelInitializer<TcpSocketChannel>(channel =>
+                {
+                    channel.Pipeline.AddLast(this.GetEncoder())
+                        .AddLast(this.GetDecoder())
+                        .AddLast(new IntCodec(true))
+                        .AddLast(new CounterHandlerInbound(this.inboundThroughputCounter))
+                        .AddLast(new CounterHandlerOutbound(this.outboundThroughputCounter))
+                        .AddLast(new ErrorCounterHandler(this.errorCounter));
+                }));
```
nayato: in the pipeline there's no final handler that would accept messages on ChannelRead and not forward them. As a result, message is forwarded to tail handler in pipeline which logs the fact that the message was not handled and calls ToString on message itself.

test/DotNetty.Transport.Tests.Performance/Utilities/IntCodec.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            if (message is IByteBuffer)
+            {
+                var buf = (IByteBuffer)message;
+                int integer = buf.ReadInt();
+                if (this.ReleaseMessages)
+                {
+                    ReferenceCountUtil.SafeRelease(message);
+                }
+                context.FireChannelRead(integer);
```
nayato: FYI, this is boxing. Not sure if it's a good thing to do in benchmark

test/DotNetty.Transport.Tests.Performance/Sockets/TcpServerSocketChannelHorizontalScaleStressSpec.cs
Aaronontheweb: So modify one of these existing handlers to accept the read and end the propagation? Or write a new one that acts as a "blackhole" of sorts?

test/DotNetty.Transport.Tests.Performance/Sockets/TcpServerSocketChannelHorizontalScaleStressSpec.cs
Aaronontheweb: Nope - but worth a try

test/DotNetty.Transport.Tests.Performance/Sockets/TcpServerSocketChannelHorizontalScaleStressSpec.cs
nayato: yep, either way would work. New one is more composable of course.

nayato:
closing as we've agreed offline that the approach for load regulation would need correction

PR #116, opened 2016-05-24
Fixes TlsHandler from operation, perf and compat POVs.

nayato:
Motivation:
Fix `TlsHandler`, optimize encryption / decryption wrt Write vs Flush semantics, optimize memory retention for sparse communication, adopt `IByteBuffer.GetIoBuffers()` in `TlsHandler`.

Modifications:
- `TlsHandler` is rewritten to use batch->copy->write approach instead of writing directly to `SslStream` in order to reduce overhead of framing / writing out everything separately.
- `TlsHandler` uses `IByteBuffer.GetIoBuffers()` instead of direct array access increasing compatibility for non-array-backed buffers.
- `TlsHandler` and `TlsHandler.MediationStream` use Task semantics now when reading through `SslStream`.
- Extra: upgraded to NBench 2.2, xUnit 2.1.0.
- Extra: better portability for .NET Core support.

Result:
`TlsHandler` provides better framing and much better (x2+) performance (depends on application control of flushing).
`TlsHandler` operates properly in edge cases when more than one frame is accepted at once and if decrypted frame size is more than 4 KB.
`TlsHandler` holds only 256 byte buffer while waiting for more data to arrive off the wire.
Extra: DotNetty is almost .NET Core compatible.

### DIFF
--- a/src/DotNetty.Handlers/Tls/TlsHandler.cs
+++ b/src/DotNetty.Handlers/Tls/TlsHandler.cs
@@ -21,23 +21,26 @@ namespace DotNetty.Handlers.Tls
 
     public sealed class TlsHandler : ByteToMessageDecoder
     {
-        const int ReadBufferSize = 4 * 1024; // todo: research perfect size
+        const int FallbackReadBufferSize = 256;
+        const int UnencryptedWriteBatchSize = 14 * 1024;
 
         static readonly Exception ChannelClosedException = new IOException("Channel is closed");
-        static readonly Action<Task, object> AuthenticationCompletionCallback = new Action<Task, object>(HandleAuthenticationCompleted);
-        static readonly AsyncCallback SslStreamReadCallback = new AsyncCallback(HandleSslStreamRead);
+        static readonly Action<Task, object> HandshakeCompletionCallback = new Action<Task, object>(HandleHandshakeCompleted);
 
         readonly SslStream sslStream;
-        State state;
+        TlsHandlerState state;
+        int packetLength;
         readonly MediationStream mediationStream;
-        IByteBuffer sslStreamReadBuffer;
         volatile IChannelHandlerContext capturedContext;
-        PendingWriteQueue pendingUnencryptedWrites;
+        BatchingPendingWriteQueue pendingUnencryptedWrites;
         Task lastContextWriteTask;
         readonly TaskCompletionSource closeFuture;
         readonly bool isServer;
         readonly X509Certificate2 certificate;
         readonly string targetHost;
+        bool firedChannelRead;
+        IByteBuffer pendingSslStreamReadBuffer;
+        Task<int> pendingSslStreamReadFuture;
 
         TlsHandler(bool isServer, X509Certificate2 certificate, string targetHost, RemoteCertificateValidationCallback certificateValidationCallback)
         {
@@ -112,46 +115,42 @@ namespace DotNetty.Handlers.Tls
             return false;
         }
 
-        static void HandleAuthenticationCompleted(Task task, object state)
+        static void HandleHandshakeCompleted(Task task, object state)
         {
             var self = (TlsHandler)state;
             switch (task.Status)
             {
                 case TaskStatus.RanToCompletion:
-                {
-                    State oldState = self.state;
-                    if ((oldState & State.AuthenticationCompleted) == 0)
                     {
-                        self.state = (oldState | State.Authenticated) & ~(State.Authenticating | State.FlushedBeforeHandshake);
+                        TlsHandlerState oldState = self.state;
+
+                        Contract.Assert(!oldState.HasAny(TlsHandlerState.AuthenticationCompleted));
+                        self.state = (oldState | TlsHandlerState.Authenticated) & ~(TlsHandlerState.Authenticating | TlsHandlerState.FlushedBeforeHandshake);
 
                         self.capturedContext.FireUserEventTriggered(TlsHandshakeCompletionEvent.Success);
 
-                        if ((oldState & State.ReadRequestedBeforeAuthenticated) == State.ReadRequestedBeforeAuthenticated
-                            && !self.capturedContext.Channel.Configuration.AutoRead)
+                        if (oldState.Has(TlsHandlerState.ReadRequestedBeforeAuthenticated) && !self.capturedContext.Channel.Configuration.AutoRead)
                         {
                             self.capturedContext.Read();
                         }
 
-                        if ((oldState & State.FlushedBeforeHandshake) != 0)
+                        if (oldState.Has(TlsHandlerState.FlushedBeforeHandshake))
                         {
                             self.Wrap(self.capturedContext);
                             self.capturedContext.Flush();
                         }
+                        break;
                     }
-                    break;
-                }
                 case TaskStatus.Canceled:
                 case TaskStatus.Faulted:
-                {
-                    // ReSharper disable once AssignNullToNotNullAttribute -- task.Exception will be present as task is faulted
-                    State oldState = self.state;
-                    if ((oldState & State.AuthenticationCompleted) == 0)
                     {
-                        self.state = (oldState | State.FailedAuthentication) & ~State.Authenticating;
+                        // ReSharper disable once AssignNullToNotNullAttribute -- task.Exception will be present as task is faulted
+                        TlsHandlerState oldState = self.state;
+                        Contract.Assert(!oldState.HasAny(TlsHandlerState.AuthenticationCompleted));
+                        self.state = (oldState | TlsHandlerState.FailedAuthentication) & ~TlsHandlerState.Authenticating;
+                        self.HandleFailure(task.Exception);
+                        break;
                     }
-                    self.HandleFailure(task.Exception);
-                    break;
-                }
                 default:
                     throw new ArgumentOutOfRangeException(nameof(task), "Unexpected task status: " + task.Status);
             }
@@ -161,7 +160,7 @@ namespace DotNetty.Handlers.Tls
         {
             base.HandlerAdded(context);
             this.capturedContext = context;
-            this.pendingUnencryptedWrites = new PendingWriteQueue(context);
+            this.pendingUnencryptedWrites = new BatchingPendingWriteQueue(context, UnencryptedWriteBatchSize);
             if (context.Channel.Active && !this.isServer)
             {
                 // todo: support delayed initialization on an existing/active channel if in client mode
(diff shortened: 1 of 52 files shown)


nayato:
@mastermanu, @mtuchkov PTAL. TlsHandler.cs is the focus (`TlsHandler.Unwrap` and `Wrap` to be precise).

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -144,10 +146,10 @@ static void HandleAuthenticationCompleted(Task task, object state)
                 case TaskStatus.Faulted:
                 {
                     // ReSharper disable once AssignNullToNotNullAttribute -- task.Exception will be present as task is faulted
-                    State oldState = self.state;
-                    if ((oldState & State.AuthenticationCompleted) == 0)
+                    HandlerState oldState = self.state;
+                    if (!oldState.Has(HandlerState.AuthenticationCompleted))
```
nayato: `.HasAny`

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -112,27 +115,26 @@ bool IgnoreException(Exception t)
...
         {
             var self = (TlsHandler)state;
             switch (task.Status)
             {
                 case TaskStatus.RanToCompletion:
                 {
-                    State oldState = self.state;
-                    if ((oldState & State.AuthenticationCompleted) == 0)
+                    HandlerState oldState = self.state;
+                    if (!oldState.Has(HandlerState.AuthenticationCompleted))
```
nayato: `.HasAny`

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -21,23 +21,26 @@ namespace DotNetty.Handlers.Tls
 
     public sealed class TlsHandler : ByteToMessageDecoder
     {
-        const int ReadBufferSize = 4 * 1024; // todo: research perfect size
+        const int FallbackReadBufferSize = 256;
```
mastermanu: [Nit] add InBytes suffix please

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -21,23 +21,26 @@ namespace DotNetty.Handlers.Tls
...
+        const int UnencryptedWriteBatchSize = 14 * 1024;
```
mastermanu: [Nit] add InBytes suffix please

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -144,10 +146,10 @@ static void HandleAuthenticationCompleted(Task task, object state)
...
                     // ReSharper disable once AssignNullToNotNullAttribute -- task.Exception will be present as task is faulted
-                    State oldState = self.state;
-                    if ((oldState & State.AuthenticationCompleted) == 0)
+                    HandlerState oldState = self.state;
+                    if (!oldState.HasAny(HandlerState.AuthenticationCompleted))
                     {
-                        self.state = (oldState | State.FailedAuthentication) & ~State.Authenticating;
+                        self.state = (oldState | HandlerState.FailedAuthentication) & ~HandlerState.Authenticating;
                     }
                     self.HandleFailure(task.Exception);
```
mastermanu: should this only be in the above if clause? what scenario breaks if we do that?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+            int endOffset = input.WriterIndex;
+            int offset = startOffset;
+            int totalLength = 0;
 
-            if (!this.EnsureAuthenticated())
+            // If we calculated the length of the current SSL record before, use that information.
+            if (this.packetLength > 0)
             {
-                return;
+                if (endOffset - startOffset < this.packetLength)
```
mastermanu: [Nit] this isn't a method or property on IByteBuffer?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+            while (totalLength < TlsUtils.MAX_ENCRYPTED_PACKET_LENGTH)
+            {
+                int readableBytes = endOffset - offset;
+                if (readableBytes < TlsUtils.SSL_RECORD_HEADER_LENGTH)
+                {
+                    break;
+                }
+
+                int packetLength = TlsUtils.GetEncryptedPacketLength(input, offset);
+                if (packetLength == -1)
```
mastermanu: i assume its not possible to read one packet and then get -1 on the subsequent call, right?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+
+                if (packetLength > readableBytes)
+                {
+                    // wait until the whole packet can be read
+                    this.packetLength = packetLength;
+                    break;
+                }
+
+                int newTotalLength = totalLength + packetLength;
+                if (newTotalLength > TlsUtils.MAX_ENCRYPTED_PACKET_LENGTH)
```
mastermanu: so let's say we read 1 packet and then try to read the 2nd packet, but it exceeds the max_encrypted_packet_length?
don't we need to update this.packetLength? Or do we do that later in this method?

src/DotNetty.Handlers/Tls/TlsHandler.cs
mastermanu: i see that we throw anyways later down the line. nevermind

---
In reply to: [64496839](https://github.com/Azure/DotNetty/pull/116#discussion_r64496839) [](ancestors = 64496839)

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+            this.firedChannelRead = false;
+            ctx.FireChannelReadComplete();
+        }
+
+        void ReadIfNeeded(IChannelHandlerContext ctx)
+        {
+            // If handshake is not finished yet, we need more data.
+            if (!ctx.Channel.Configuration.AutoRead && (!this.firedChannelRead || !this.state.HasAny(HandlerState.AuthenticationCompleted)))
+            {
+                // No auto-read used and no message passed through the ChannelPipeline or the handhshake was not complete
```
mastermanu: [Nit] spelling

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -619,4 +858,25 @@ public WaitHandle AsyncWaitHandle
...
+        AuthenticationCompleted = Authenticated | FailedAuthentication
+    }
+
+    static class HandlerStateExtensions
+    {
+        public static bool Has(this HandlerState value, HandlerState testValue) => (value & testValue) == testValue;
+
+        public static bool HasAny(this HandlerState value, HandlerState testValue) => (value & testValue) != 0;
+
+        public static HandlerState Sans(this HandlerState value, HandlerState exclusionValue) => value & ~exclusionValue;
```
mastermanu: used anywhere?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -112,27 +115,26 @@ bool IgnoreException(Exception t)
...
         {
             var self = (TlsHandler)state;
             switch (task.Status)
             {
                 case TaskStatus.RanToCompletion:
                 {
-                    State oldState = self.state;
-                    if ((oldState & State.AuthenticationCompleted) == 0)
+                    HandlerState oldState = self.state;
+                    if (!oldState.HasAny(HandlerState.AuthenticationCompleted))
```
mastermanu: remove if checks and replace with asserts?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+        void Unwrap(IChannelHandlerContext ctx, IByteBuffer packet, int offset, int length, List<object> output)
+        {
+            //bool notifyClosure = false; // todo: netty/issues/137
+            bool pending = false;
+
+            IByteBuffer decodeOut = null;
+
+            try
+            {
+                this.mediationStream.AcceptBytes(packet.GetIoBuffer(offset, length));
```
mastermanu: rename...AcceptBytes implies data getting *returned*

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+
+                if (!this.EnsureAuthenticated())
+                {
+                    return;
+                }
+
+                Task<int> currentReadFuture = this.pendingReadFuture;
```
mastermanu: this.pendingSslStreamReadFuture

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+
+                if (currentReadFuture != null && !currentReadFuture.IsCompleted)
```
mastermanu: Assert

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+                {
+                    return;
+                }
+
+                decodeOut = this.pendingReadBuffer ?? ctx.Allocator.Buffer(length);
```
mastermanu: pendingSslStreamReadBuffer

mastermanu: combine with above assert (someehow)

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+                    {
+                        int readInstant = currentReadFuture.Result;
+
+                        Contract.Assert(readInstant > 0);
+
+                        outletOffset += readInstant;
+                        read += readInstant;
+
+                        IByteBuffer swappedOut;
+                        int remainingInput = this.mediationStream.SourceLength; // bytes not yet read by SslStream
```
mastermanu: rename SourceLength

src/DotNetty.Handlers/Tls/TlsHandler.cs
mastermanu: add comment

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+                            else
+                            {
+                                // current output buffer is full - very likely there's more data in SslStream's internal state
+                                decodeOut = ctx.Allocator.Buffer(Math.Max(FallbackReadBufferSize, length - read));
+                            }
+                        }
+                        else
+                        {
+                            // SslStream has more input to go through - it won't require more input on next ReadAsync call
+                            if (outlet.Count - outletOffset >= remainingInput)
```
mastermanu: check if decoding 2nd packet can result in 0 byte read from SSL Stream perspective. if so, need to add to output

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+                            if (outletOffset < outlet.Count)
+                            {
+                                // current output buffer isn't full - SslStream will likely block =>
+                                // release current result and switch to buffer of FallbackReadBufferSize
+                                decodeOut = ctx.Allocator.Buffer(FallbackReadBufferSize);
+                            }
+                            else
+                            {
+                                // current output buffer is full - very likely there's more data in SslStream's internal state
+                                decodeOut = ctx.Allocator.Buffer(Math.Max(FallbackReadBufferSize, length - read));
```
mastermanu: why not pass remainingInput here?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +182,257 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+                            if (outlet.Count - outletOffset >= remainingInput)
+                            {
+                                // use the same output buffer
+                                swappedOut = null;
+                            }
+                            else
+                            {
+                                // release the buffer and provide another sufficient output buffer
+                                swappedOut = decodeOut;
+                                decodeOut = ctx.Allocator.Buffer(length - read);
```
mastermanu: remainingInput?

src/DotNetty.Handlers/Tls/TlsHandler.cs
mastermanu: what's the heuristic here? technically we've exhausted all the remaining input? how much do we expect this to be?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -21,23 +21,26 @@ namespace DotNetty.Handlers.Tls
...
     public sealed class TlsHandler : ByteToMessageDecoder
     {
-        const int ReadBufferSize = 4 * 1024; // todo: research perfect size
+        const int FallbackReadBufferSize = 256;
+        const int UnencryptedWriteBatchSize = 14 * 1024;
 
         static readonly Exception ChannelClosedException = new IOException("Channel is closed");
-        static readonly Action<Task, object> AuthenticationCompletionCallback = new Action<Task, object>(HandleAuthenticationCompleted);
-        static readonly AsyncCallback SslStreamReadCallback = new AsyncCallback(HandleSslStreamRead);
+        static readonly Action<Task, object> AuthenticationCompletionCallback = new Action<Task, object>(HandleHandshakeCompleted);
```
mastermanu: [Nit] HandshakeCompletionCallback

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +179,306 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+            int offset = startOffset;
+            int totalLength = 0;
 
-            if (!this.EnsureAuthenticated())
+            List<int> packetLengths;
+            // if we calculated the length of the current SSL record before, use that information.
+            if (this.packetLength > 0)
             {
-                return;
+                if (endOffset - startOffset < this.packetLength)
```
mastermanu: [Nit] add comment indicating that there is still not enough data left to read through the remaining packet.

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +179,306 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+                Contract.Assert(packetLength > 0);
+
+                if (packetLength > readableBytes)
+                {
+                    // wait until the whole packet can be read
+                    this.packetLength = packetLength;
+                    break;
+                }
+
+                int newTotalLength = totalLength + packetLength;
```
mastermanu: btw i assume this also includes the bytes involved in encoding the length?

the comment for GetEncryptedPacketLength says that the reader index is not incremented

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +179,306 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+                    packetLengths = new List<int>(4);
+                    packetLengths.Add(this.packetLength);
+                    offset += this.packetLength;
+                    totalLength = this.packetLength;
+                    this.packetLength = 0;
+                }
+            }
+            else
+            {
+                packetLengths = new List<int>(4);
```
mastermanu: btw, I believe this is the default length of an initialized List.

Also, can we recycle / reuse this across calls?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +179,306 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+            {
+                ArraySegment<byte> inputIoBuffer = packet.GetIoBuffer(offset, length);
+                this.mediationStream.SetSource(inputIoBuffer.Array, inputIoBuffer.Offset);
+
+                int packetIndex = 0;
+
+                while (!this.EnsureAuthenticated())
+                {
+                    this.mediationStream.ExpandSource(packetLengths[packetIndex]);
+                    if (++packetIndex >= packetLengths.Count)
```
mastermanu: [Nit] ==

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -180,47 +179,306 @@ protected override void HandlerRemovedInternal(IChannelHandlerContext context)
...
+                            // we did feed the whole current packet to SslStream yet it did not produce any result -> move to the next packet in input
+                            Contract.Assert(this.mediationStream.SourceReadableBytes == 0);
+
+                            continue;
+                        }
+
+                        int read = currentReadFuture.Result;
+
+                        // Now output the result of previous read and decide whether to do an extra read on the same source or move forward
+                        AddBufferToOutput(outputBuffer, read, output);
```
mastermanu: [Nit] naming

src/DotNetty.Handlers/Tls/TlsHandler.cs
nayato: default is `0`. We can see if there's visible benefit in pooling. it's different than `List<object>` that we're using in ByteToMessageDecoder as it is more specialized / not as chatty in generic case.

src/DotNetty.Handlers/Tls/TlsHandler.cs
nayato: that's correct

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -280,61 +521,93 @@ public override void Flush(IChannelHandlerContext context)
...
 
-                var buf = (IByteBuffer)msg;
-                buf.ReadBytes(this.sslStream, buf.ReadableBytes); // this leads to FinishWrap being called 0+ times
+                    if (messages.Count == 1)
+                    {
+                        buf = (IByteBuffer)messages[0];
+                    }
+                    else
+                    {
+                        buf = context.Allocator.Buffer((int)this.pendingUnencryptedWrites.CurrentSize);
```
mtuchkov: Why do we need buf? why can't we write to sslStream directly?

mastermanu:
:shipit:

src/DotNetty.Handlers/Tls/TlsHandler.cs
nayato: not to package every single buffer as a separate SSL record. Every call to `SslStream.Write(...)` results in a new SSL record.

PR #121, opened 2016-06-01
Check MQTT packet identifier > 0. Fixes #76

### LINKED ISSUE #76: MQTT Decoder does not have validation for Packet ID "0" and is currently allowed.
Shouldn't  we have a validation in MQTT decoder for validating the Packet ID that it should always be greater than "0" ?. I see a method in utility for validating this but not referenced.

nayato:


nayato:
@mtuchkov PTAL

mtuchkov:
LGTM

PR #134, opened 2016-06-18
Negotiage client certificate

mtuchkov:
Motivation:
SslStream's paramter required client certificate is misleading. In fact in suggests to client to provide the certificate and client may or may not provide it.
If client has not provided the client certificate - it assumes that client is not authenticated from TLS prospective and continues the handshake

Modification:
Expose this setting as a parameter, so that users could set it if needed.

Results:
Allows users to do a client certificate authentication on server.


nayato:
LGTM

Issue #133, opened 2016-06-16
Object Disposed Exception running EchoClient and EchoServer


GrahamDarnell:
Run the EchoServer, run EchoClient, hit enter on EchoClient to close it down. The EchoServer throws ObjectDisposedException when attempting this.Socket.Receive() in TcpSocketChannel::DoReadBytes()

nayato:
Known issue, will be fixed in an upcoming release. Also applies to `TlsHandler`.

PR #152, opened 2016-08-07
Fixes excessive exceptions from being thrown in case of graceful chan…

nayato:
…nel closure.

Motivation:
SocketException and ODE are being thrown when disconnecting gracefully.

Modifications:
- dismiss async read completion if it is done on inactive channel
- do not attempt sync-reading from a socket that is not Connected
- extra: fixing echo sample with better practices

Result:
No excessive errors happen during graceful shutdown


nayato:
@mtuchkov, PTAL

nayato:
Fixes #133

src/DotNetty.Transport/Channels/Sockets/AbstractSocketChannel.cs
```diff
@@ -108,7 +108,18 @@ protected internal void ClearReadPending()
 
         protected void SetState(StateFlags stateToSet) => this.state |= stateToSet;
 
-        protected bool ResetState(StateFlags stateToReset)
+        /// <returns>state before modification</returns>
+        protected StateFlags ResetState(StateFlags stateToReset)
+        {
+            StateFlags oldState = this.state;
+            if ((oldState & stateToReset) != 0)
```
mtuchkov: Is it possible that the old state is A&B and a new state is B&C ?

src/DotNetty.Transport/Channels/Sockets/TcpServerSocketChannel.cs
```diff
@@ -127,10 +127,13 @@ public TcpServerSocketChannelUnsafe(TcpServerSocketChannel channel)
 
             public override void FinishRead(SocketChannelAsyncOperation operation)
             {
-                Contract.Requires(this.channel.EventLoop.InEventLoop);
+                Contract.Assert(this.channel.EventLoop.InEventLoop);
 
                 TcpServerSocketChannel ch = this.Channel;
-                ch.ResetState(StateFlags.ReadScheduled);
+                if ((ch.ResetState(StateFlags.ReadScheduled) & StateFlags.Active) == 0)
```
mtuchkov: can we come here when we already are in the ReadScheduled state?

src/DotNetty.Transport/Channels/Sockets/TcpSocketChannel.cs
```diff
@@ -164,12 +164,18 @@ void OnConnected()
 
         protected override void DoClose()
         {
-            base.DoClose();
-            if (this.ResetState(StateFlags.Open | StateFlags.Active))
+            try
             {
-                this.Socket.Shutdown(SocketShutdown.Both);
-                this.Socket.Dispose();
+                if (this.TryResetState(StateFlags.Open | StateFlags.Active))
```
mtuchkov: On DoClose we try to set the Open&Active state. Is this correct?

src/DotNetty.Transport/Channels/Sockets/TcpServerSocketChannel.cs
nayato: 100% ch is in ReadScheduled state.

---
In reply to: [73931101](https://github.com/Azure/DotNetty/pull/152#discussion_r73931101) [](ancestors = 73931101)

mtuchkov:
:shipit:

src/DotNetty.Transport/Channels/Sockets/TcpServerSocketChannel.cs
mtuchkov: OK

PR #156, opened 2016-08-19
Fixes TLS initialization

nayato:
Motivation:
TLS start got broken in one of last commits: just connecting to server does not trigger handshake anymore.

Modifications:
reverted server vs client behavior decisions from previous commits.

Result:
Client-side TLS handler will start the handshake once underlying connection is established.


nayato:
@mtuchkov PTAL

mtuchkov:
:shipit:

PR #159, opened 2016-09-03
Allow looking up current executor

nayato:
Motivation:
In many instances routines may be optimized in a presence of IEventExecutor but it is hard to carry it around everywhere.

Modifications:
- introduced ExecutionEnvironment.TryGetCurrentExecutor which looks up current executor if any.
- modified SingleThreadEventExecutor to set itself as current on its worker thread.

Result:
Executor may be looked up ambiently for optional executor-specific optimizations.


mastermanu:
LGTM

PR #160, opened 2016-09-06
Switch to ConcurrentQueue in STEE by default

nayato:
Motivation:
as netty moved on to use JCTools directly, it is worth revisiting standard .NET concurrent collections and open queue type election in executor.

Modifications:
- Removed MpscLinkedQueue as redundant (perf is no better than with ConcurrentQueue)
- Introduced CompatibleConcurrentQueue to bridge ConcurrentQueue and IQueue<T> interface, made it the default.
- Updated STEE execution benchmark

Result:
Cleaner solution, less GC pressure from executor actions, ability to specify specialized task queue type (e.g. MpscArrayQueue on server for efficiency (at a cost of extra-mem consumption).


nayato:
@mtuchkov PTAL

nayato:
FYI, creating a custom concurrent collection akin ConcurrentQueue doesn't help much. Might need to port MpscChunkedArrayQueue to get better results.

src/DotNetty.Common/Concurrency/SingleThreadEventExecutor.cs
```diff
@@ -41,6 +41,7 @@ public class SingleThreadEventExecutor : AbstractScheduledEventExecutor
         public SingleThreadEventExecutor(string threadName, TimeSpan breakoutInterval)
         {
             this.terminationCompletionSource = new TaskCompletionSource();
+            this.taskQueue = this.NewTaskQueue();
```
mtuchkov: A virtual method call in ctor. if you are 100% sure - let's add some comment then

src/DotNetty.Common/Concurrency/SingleThreadEventExecutor.cs
```diff
@@ -104,7 +107,7 @@ public override void Execute(IRunnable task)
 
         protected void WakeUp(bool inEventLoop)
         {
-            if (!inEventLoop || this.executionState == ST_SHUTTING_DOWN)
+            if (!inEventLoop || (this.executionState == ST_SHUTTING_DOWN))
```
mtuchkov: aren't the parenthesis redundant? if that is just for clarity - I'm fine.

nayato: I think that's ReSharper's new auto-format adding them for clarity..

src/DotNetty.Common/Concurrency/SingleThreadEventExecutor.cs
```diff
@@ -356,33 +359,32 @@ bool RunAllTasks(PreciseTimeSpan timeout)
...
                 {
-                    IScheduledRunnable scheduledTask = this.PollScheduledTask(nanoTime);
-                    if (scheduledTask == null)
-                    {
-                        break;
-                    }
-
-                    this.taskQueue.TryEnqueue(scheduledTask);
+                    // No space left in the task queue add it back to the scheduledTaskQueue so we pick it up again.
+                    this.ScheduledTaskQueue.Enqueue(scheduledTask);
```
mtuchkov: Is that OK that we take it from the head of the scheduledTaskQueue add put it back to the tail?

mtuchkov: Although, that should be a rare scenario, but it can lead to timeout for this particular request that is being processed.  Isn't it?

nayato: `scheduledTaskQueue` is a priority queue. While reordering is possible amid items with the same time (down to ticks), users should not rely on this queue for ordering really.

src/DotNetty.Common/Concurrency/SingleThreadEventExecutor.cs
nayato: thinking about it again, I'll add a protected ctor instead

src/DotNetty.Common/Internal/RefArrayAccessUtil.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    static class RefArrayAccessUtil
+    {
+        public static readonly int RefBufferPad = 64 * 2 / IntPtr.Size;
+
+        /// A plain store (no ordering/fences) of an element to a given offset
+        /// @param buffer this.buffer
+        /// @param offset computed via {@link ConcurrentCircularArrayQueue#calcElementOffset(long)}
+        /// @param e an orderly kitty
+        public static void SpElement<T>(T[] buffer, long offset, T e) => buffer[offset] = e;
```
mtuchkov: Is it really handy or this is just for porting purposes?

mtuchkov:
:shipit:

src/DotNetty.Common/Internal/RefArrayAccessUtil.cs
nayato: alignment in port

PR #170, opened 2016-11-04
Move to .NET Core tooling

nayato:
Motivation
.NET Core tooling allows building for both .NET Framework and .NET Standard while easing pains of running tests and packaging.

Modifications
- csproj -> xproj + project.json
- support for .NET Standard 1.3
- build script is now based on CAKE
- using dotnet CLI for building, testing, packing
- strong name by default (snk included)

Result
DotNetty works on .NET Core and any other platform supporting .NET Standard 1.3+.


nayato:
@mtuchkov PTAL

nayato:
FYI @davidfowl @terrajobst

examples/Factorial/project.json
```diff
@@ -0,0 +1,29 @@
...
+  "frameworks": {
+    "netcoreapp1.0": {
+      "dependencies": {
+        "Microsoft.NETCore.App": {
+          "version": "1.0.0-*",
+          "type": "platform"
+        }
+      },
+      "imports": "dnxcore50"
+    }
```
nayato: I couldn't figure out how to make `net451` work here due to `BigInteger`. Adding dependency on `System.Numerics` won't work, with dependency on `System.Runtime.Numerics` still fails to build `net451`. cc @terrajobst,

alexvaluyskiy:
@nayato maybe you should wait when MS will switch from project.json to MSBuild?

examples/Factorial/project.json
alexvaluyskiy: Have you tried frameworkAssemblies?
```json
    "net451": {
      "frameworkAssemblies": {
        "System.Numerics": "4.0.0.0"
      }
    } 
```

nayato:
@alexvaluyskiy that would be much easier change. Besides, they promise a migration tool :)
With .NET Core picking up now, doesn't make sense to wait more (another 2+ months?) until there is a stable tooling release.

davidfowl:
:shipit:

Issue #79, opened 2016-03-23
Implement Socket-based UDP channel


nayato:

startemplar:
Support for Socket-based UDP channel in AbstractSocketChannel.OnIoCompleted

The method is called after an async socket operation completed. Currently the SocketAsyncOperation.Receive, SocketAsyncOperation.Send for TCP. However, UDP uses ReceiveFrom, SendTo pattern
so we need to add support here, like
case SocketAsyncOperation.ReceiveFrom:
case SocketAsyncOperation.SendTo:

Thank for your comment/suggestions.

StormHub:
I am porting/re-implementing UDP channel.

nayato:
@StormHub you may be interested in @onatm's progress on this: https://github.com/onatm/DotNetty/tree/implement-socket-based-udp-channel

onatm:
@nayato @StormHub i couldn't make so much progress since i started but i'll be happy to help

StormHub:
It's ok. With UDP I almost have exactly the same functions as original Netty :)

StormHub:
Current progress on https://github.com/StormHub/DotNetty/tree/UDP
Should I create some benchmark on the channel?

StormHub:
Initial implementation is roughly done now.

StormHub:
@nayato probably start another PR for this soon. I am running UDP channel in real world and how it goes.

StormHub:
Sorry to add more extra work load for you :)

StormHub:
@nayato DatagramChannel final change set before PR. Anything you see I probably missed?

StormHub:
PR  #171 created, please review.

PR #171, opened 2016-11-06
Socket-based UDP channel transport (#79).

StormHub:
Motivation:
Ported/Re-implemented DatagramChannel/UDP into DotNetty

Modifications:
- Added DatagramChannel/Config into transport project.
- Added DatagramPacket/DefaultAddressedEnvelop/IAddressedEnvelop into transport project.
- Added DatagramPacketEncoder/DatagramPacketDecoder into Codec project.
- Extra: Ported QuoteOfTheMoment UDP example project.

Result:
DatagramChannel/UDP transport is supported by DotNetty.


azurecla:
Hi __@StormHub__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <span>We will now validate the agreement and then real humans will evaluate your PR.</span>

TTYL, AZPRBOT;

Aaronontheweb:
@StormHub the UDP multi-cast tests hung the build on two separate attempts. Can you reproduce that locally?

StormHub:
@Aaronontheweb Yep, I noticed it. Just added a logger to catch what happened.

StormHub:
@Aaronontheweb I won't push again until i get the instrumentation right locally.

nayato:
@StormHub can you please add timeouts in tests to avoid hangs in the future (`DotNetty.Tests.Common.TaskExtensions.WithTimeout(..)` might help).

StormHub:
@nayato already did. The problem was that multicast join group.

nayato:
@StormHub, as a sanity check, could you please add a benchmark to DotNetty.Transport.Tests.Performance/Sockets based on `TcpChannelPerfSpecs` using UDP in place of TCP?

StormHub:
@nayato There is a benchmark spec DatagramChannelPerfSpecs,
"Measures average DatagramChannel overhead of broadcasting 1000 64 byte messages from client to server."

StormHub:
@nayato Now both UDP unicast and multicast pass. Previously xunit ran both in parallel causing each other to time out on binding to local loopback address. I have added a collection of "UDP Transport Tests" with both to prevent it from happening. It works now :)

src/DotNetty.Codecs/Base64/Base64Decoder.cs
```diff
@@ -3,6 +3,7 @@
```
nayato: what are these changes about? Formatting before actually looks better than suggested now.

StormHub: Might be merge changes. I reformatted it.

src/DotNetty.Transport/Channels/Sockets/DatagramChannel.cs
```diff
@@ -0,0 +1,460 @@
...
+
+        protected override void ScheduleSocketRead()
+        {
+            SocketChannelAsyncOperation operation = this.ReadOperation;
+            operation.RemoteEndPoint = new IPEndPoint(
+                this.LocalAddress.AddressFamily == AddressFamily.InterNetwork ? IPAddress.Any : IPAddress.IPv6Any,
+                IPEndPoint.MinPort);
+
+            IRecvByteBufAllocatorHandle handle = this.Unsafe.RecvBufAllocHandle;
+            IByteBuffer buffer = handle.Allocate(this.config.Allocator);
```
nayato: have you tried zero byte buffer like in `AbstractSocketByteChannel`? Does it never complete or does it return a specific error code once message is available but can't be read into the buffer (due to a size issue)?

nayato: What happens if message does not fit in the buffer? Will it be discarded? Will we receive partial message? Pls check and also see if MSG_PARTIAL (https://msdn.microsoft.com/en-us/library/windows/desktop/ms741686(v=vs.85).aspx) can be used here to have more flexibility in terms of buffer management.

src/DotNetty.Transport/Channels/Sockets/DatagramChannel.cs
```diff
@@ -0,0 +1,460 @@
...
+            handle.AttemptedBytesRead = buffer.WritableBytes;
+            operation.UserToken = buffer;
+
+            ArraySegment<byte> bytes = buffer.GetIoBuffer();
+            operation.SetBuffer(bytes.Array, bytes.Offset, buffer.WritableBytes);
+            bool pending = this.Socket.ReceiveFromAsync(operation);
```
nayato: I guess it can benefit from `ExecutionContext.SuppressFlow()`. Pls check `AbstractSocketByteChannel`'s version for details. That avoids execution context copy on `ReceiveAsync` call.

src/DotNetty.Transport/Channels/Sockets/DatagramChannel.cs
```diff
@@ -0,0 +1,460 @@
...
+
+                operation.UserToken = null;
+            }
+        }
+
+        static void OnReceiveFromCompletedSync(object u, object p) => ((ISocketChannelUnsafe)u).FinishRead((SocketChannelAsyncOperation)p);
+
+        protected override void ScheduleMessageWrite(object message)
+        {
+            var envolope = message as IAddressedEnvelope<IByteBuffer>;
```
nayato: typo, pls check elsewhere

src/DotNetty.Transport/Channels/Sockets/DatagramChannel.cs
```diff
@@ -0,0 +1,460 @@
...
+            }
+
+            int length = data.ReadableBytes;
+            if (length == 0)
+            {
+                return true;
+            }
+
+            ArraySegment<byte> bytes = data.GetIoBuffer(data.ReaderIndex, length);
+            int writtenBytes = this.Socket.SendTo(bytes.Array, data.ArrayOffset + data.ReaderIndex, length, SocketFlags.None, remoteAddress);
```
nayato: use `bytes.Offset` for offset and `bytes.Count` for count here.

nayato:
@StormHub I've managed to look through the first half of this. I'll try to get through the rest tomorrow EOD (PST).

src/DotNetty.Codecs/Base64/Base64Decoder.cs
```diff
@@ -21,7 +21,7 @@ public Base64Decoder(Base64Dialect dialect)
             this.dialect = dialect;
         }
 
-        protected override void Decode(IChannelHandlerContext context, IByteBuffer message, List<object> output) => output.Add(Base64.Decode(message, this.dialect));
+        protected internal override void Decode(IChannelHandlerContext context, IByteBuffer message, List<object> output) => output.Add(Base64.Decode(message, this.dialect));
```
StormHub: It is for https://github.com/StormHub/DotNetty/blob/UDP/src/DotNetty.Codecs/DatagramPacketEncoder.cs

src/DotNetty.Codecs/MessageToMessageEncoder.cs
```diff
@@ -101,6 +101,6 @@ public override Task WriteAsync(IChannelHandlerContext ctx, object msg)
         ///     needs to do some kind of aggragation
         ///     @throws Exception    is thrown if an error accour
         /// </summary>
-        protected abstract void Encode(IChannelHandlerContext context, T message, List<object> output);
+        protected internal abstract void Encode(IChannelHandlerContext context, T message, List<object> output);
```
StormHub: It is for the envelop encoder/decoder wrapper for IAddressedEnvelope<T>

src/DotNetty.Transport/Channels/Sockets/DatagramChannel.cs
StormHub: If the buffer size is not enough, the packet is discarded. No specific error is given.

StormHub: I believe the default maximum for UDP is 65536. It will be discarded since UDP does not offer flow control like TCP. What is the best way to handle this?

StormHub: The .NET wrapper UdpClient simply sets the buffer to be the maximum.

StormHub: The default config uses a fixed length byte buffer (default to 2048) and it is configurable.

src/DotNetty.Transport/Channels/Sockets/ISocketDatagramChannel.cs
```diff
@@ -0,0 +1,39 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace DotNetty.Transport.Channels.Sockets
+{
+    using System.Net;
+    using System.Net.NetworkInformation;
+    using System.Threading.Tasks;
+    using DotNetty.Common.Concurrency;
+
+    public interface ISocketDatagramChannel : ISocketChannel
```
nayato: My comment was about class only :)
`ISocketChannel` should be replaced with `IChannel`, ISocketDatagramChannel` -> `IDatagramChannel`.

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
```diff
@@ -0,0 +1,479 @@
...
+                data.SetWriterIndex(data.WriterIndex + received);
+                EndPoint remoteAddress = operation.RemoteEndPoint;
+                buf.Add(new DatagramPacket(data, remoteAddress, this.LocalAddress));
+                free = false;
+
+                return 1;
+            }
+            catch (Exception exception)
+            {
+                Logger.Error(exception);
```
nayato: `PlatformDependent.throwException(cause);` in netty here will actually just throw. It's a way for them to avoid specifying `throws ...` on the method while it may actually throw.
Just remove the whole `catch` block pls.

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
```diff
@@ -0,0 +1,479 @@
...
+        public SocketDatagramChannel(Socket socket)
+            : base(null, socket)
+        {
+            this.config = new DefaultSocketDatagramChannelConfig(this, socket);
+            this.anyRemoteEndPoint = new IPEndPoint(
+                socket.AddressFamily == AddressFamily.InterNetwork ? IPAddress.Any : IPAddress.IPv6Any,
+                IPEndPoint.MinPort);
+
+            // Read is not allowed until socket is bound for UDP
+            this.config.AutoRead = false;
```
nayato: AutoRead is meant for user of DotNetty to control channel's behavior when it comes to reading data out of source. Here we take control away from the user. How is it handled in netty?

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
```diff
@@ -0,0 +1,479 @@
...
+                return;
+            }
+
+            SocketChannelAsyncOperation operation = this.PrepareWriteOperation(data.GetIoBuffer(data.ReaderIndex, length));
+            operation.RemoteEndPoint = envelope.Recipient;
+            this.SetState(StateFlags.WriteScheduled);
+            bool pending = this.Socket.SendToAsync(operation);
+            if (!pending)
+            {
+               ((ISocketChannelUnsafe)this.Unsafe).FinishWrite(operation);
```
nayato: could this cause stack overflow if there is constant stream of messages and SendToAsync would complete synchronously every time?

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
StormHub: Well, in the current DotNetty, the read operation will be immediately executed once instance is created, but in UDP you cannot read until bind operation finishes. In original netty, the read block is handled by  java specific DatagramChannel itself.  What is the intended way of handling this in DotNetty.

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
StormHub: The perf test does that actually. It does not look like a problem but let me double check everything.

StormHub: I will confirm this on my mac. Put some conditionals if required.

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
nayato: um.. for TCP read starts once socket is bound / connected. Where do you see read starting earlier?

nayato: AutoRead-based reading starts in DefaultChannelPipeline.FireChannelActive. It is called in three cases:
- channel bind completes and it was not Active before completion.
- channel connect completes and it was not Active before completion
- channel is registered with executor and is already active.

I wonder if it kicks off earlier than you expect because it becomes Active prematurely per current implementation/

src/DotNetty.Transport/Channels/Sockets/DefaultDatagramChannelConfig.cs
```diff
@@ -0,0 +1,514 @@
...
+        {
+            set
+            {
+                if (this.Channel.Registered)
+                {
+                    throw new InvalidOperationException("Can only changed before channel was registered");
+                }
+                this.activeOnOpen = value;
+            }
+            get
```
nayato: >get [](start = 12, length = 3)

code style: get first. Also, pls put get on one line

src/DotNetty.Transport/Channels/Sockets/DefaultDatagramChannelConfig.cs
```diff
@@ -0,0 +1,514 @@
...
+            return true;
+        }
+
+        public bool ActiveOnOpen
+        {
+            set
+            {
+                if (this.Channel.Registered)
+                {
+                    throw new InvalidOperationException("Can only changed before channel was registered");
```
nayato: >Can only changed [](start = 57, length = 16)

Can only be changed

src/DotNetty.Transport/Channels/Sockets/DefaultDatagramChannelConfig.cs
```diff
@@ -0,0 +1,514 @@
...
+            }
+            else
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        public bool ActiveOnOpen
```
nayato: It is deprecated in netty. Can we skip it?

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
```diff
@@ -0,0 +1,470 @@
...
+
+        protected override void DoBind(EndPoint localAddress)
+        {
+            this.Socket.Bind(localAddress);
+            this.CacheLocalAddress();
+
+            this.SetState(StateFlags.Active);
+        }
+
+        public override bool Active =>
```
nayato: Open is true from object creation, Registered will become true upon call to AbstractChannel.Register0 and therefore it will call IChannelPipeline.FireChannelActive() which it turn will start reading. I am not sure how this works in netty, but it appears to me that for this implementation we have to assume ChannelActiveOnOpen being false, i.e. Active should look like so:
``` bool Active => this.Open && this.Socket.IsBound;```

Comment (StormHub) [inline src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs]: Then this is the flip side of the problem, if the channel is not active, writes will fail

Comment (StormHub) [inline src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs]: Personally I feel like keep the channel active for UDP upon register and start read on bind only.

Comment (nayato) [inline src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs]: how?

Comment (nayato) [inline src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs]: from commit in netty adding "active_on_registration" option:
> Introduce a new ChannelOption called DATAGRAM_CHANNEL_ACTIVE_ON_REGISTRATION. Related to [#1830]  This ChannelOption allows to tell the DatagramChannel implementation to be active as soon as they are registrated to their EventLoop. This can be used to make it possible to write to a not bound DatagramChannel. The ChannelOption is marked as @deprecated as I'm looking for a better solution in master which breaks default behaviour with 4.0 branch.```

Comment (StormHub) [inline src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs]: They running into the same problem :) therefore this flag was added to solve it (really funny).
Have a look at the failed unit tests. If the channel active returns false, writes will throw.

Comment (StormHub) [inline src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs]: For tcp this is perfectly valid but not really applicable to udp.

Comment (nayato) [inline src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs]: it basically comes down to a fact that for connectionless protocols there is "ready" and there is "connected" states. Former allows writes, latter allows reads. Thinking how to fit it in right now..

Comment (nayato) [inline src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs]: I'd really want to take UDP in this release.

Comment (StormHub) [inline src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs]: Me too :) I have added a CanWrite into the AbstractUnsafe, please have a look at it and give me some feedback.

File: src/DotNetty.Transport/Channels/AbstractChannel.cs [hunk 16]
```diff
@@ -738,7 +738,7 @@ protected virtual void Flush0()
                 this.inFlush0 = true;
 
                 // Mark all pending write requests as failure if the channel is inactive.
-                if (!this.channel.Active)
+                if (!this.CanWrite)
```
nayato: seems reasonable

nayato: that's just enough. If we see a bigger pattern in the future we'll rethink. For now it fits the bill.

nayato:
I'll have another sanity pass now and will likely merge if nothing else comes up. 0.4.0 is slated for EOW release.

src/DotNetty.Transport/Channels/Sockets/IDatagramChannel.cs
```diff
@@ -0,0 +1,39 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace DotNetty.Transport.Channels.Sockets
+{
+    using System.Net;
+    using System.Net.NetworkInformation;
+    using System.Threading.Tasks;
+    using DotNetty.Common.Concurrency;
+
+    public interface IDatagramChannel : ISocketChannel
```
nayato: should be `IChannel`, not `ISocketChannel`

src/DotNetty.Transport/Channels/Sockets/DefaultAddressedEnvelope.cs
```diff
@@ -0,0 +1,76 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace DotNetty.Transport.Channels.Sockets
+{
+    using System.Diagnostics.Contracts;
+    using System.Net;
+    using DotNetty.Common;
+    using DotNetty.Common.Utilities;
+
+    public class DefaultAddressedEnvelope<T> : IAddressedEnvelope<T>
```
nayato: pls pull it out to `DotNetty.Transport.Channels`

src/DotNetty.Transport/Channels/Sockets/IAddressedEnvelope.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace DotNetty.Transport.Channels.Sockets
+{
+    using System.Net;
+    using DotNetty.Common;
+
+    public interface IAddressedEnvelope<out T> : IReferenceCounted
```
nayato: pls pull it out into `DotNetty.Transport.Channels`

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
nayato: I've checked with .NET Core developers and it's the same way on Linux as of today but I'll post a fix to .NET Core so it will become a concern going forward.

nayato:
@StormHub, CI build has failed. Could you please check why?

StormHub:
Yeah, I saw that.

StormHub:
Fixed it was the wrong interface type.

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
```diff
@@ -0,0 +1,481 @@
...
+            SocketChannelAsyncOperation operation = this.ReadOperation;
+            operation.RemoteEndPoint = this.anyRemoteEndPoint;
+
+            IRecvByteBufAllocatorHandle handle = this.Unsafe.RecvBufAllocHandle;
+            IByteBuffer buffer = handle.Allocate(this.config.Allocator);
+            handle.AttemptedBytesRead = buffer.WritableBytes;
+            operation.UserToken = buffer;
+
+            ArraySegment<byte> bytes = buffer.GetIoBuffer();
+            operation.SetBuffer(bytes.Array, bytes.Offset, buffer.WritableBytes);
```
nayato: technically this is not allowed: GetIoBuffer without parameters requests for ArraySegment that is `ReadableBytes` long. pls change to `GetIoBuffer(0, buffer.WritableBytes)` and use `bytes.Count` for length.

nayato:
@StormHub, merging at long last. Thanks for patience and all the hard work that went into this PR!

StormHub:
@nayato 💃

PR #175, opened 2016-11-24
Tweaks

nayato:


nayato:
@mtuchkov PTAL

src/DotNetty.Common/Utilities/ReferenceCountUtil.cs
```diff
@@ -178,9 +202,6 @@ public static T ReleaseLater<T>(T msg, int decrement)
         }
 
         static string FormatReleaseString(IReferenceCounted referenceCounted, int decrement)
-        {
-            return referenceCounted.GetType().Name + ".Release(" + decrement.ToString() + ") refCnt: "
-                + referenceCounted.ReferenceCount.ToString();
-        }
+            => $"{referenceCounted.GetType().Name}.Release({decrement.ToString()}) refCnt: {referenceCounted.ReferenceCount.ToString()}";
```
mtuchkov: you was unhappy with string.format vs string.concat

PR #189, opened 2017-01-31
new AffinitizedEventLoopGroup; Refining EE/G and EL/G API and docs

nayato:
Motivation:
In proxy scenarios it is often important to avoid unnecessary switching between event loops while passing traffic from client and server and back.

Modifications:
- Added AffinitizedEventLoopGroup
- introduced IEventExecutor.Parent and IEventLoop.Parent
- better xml-docs for executors and event loops.

Result:
Proxy scenario is better supported out of the box.


nayato:
@myagley PTAL

PR #214, opened 2017-03-21
SNI support

krish-gh:
This is to add support for SNI extension in TLS. Right now, there is no clean solution in C# due to the lack of support in SslStream. So, one way is to add a peek stream logic to read the ClientHello for server name extension.

N.B. Credit goes to http://stackoverflow.com/questions/2196767/c-implementing-networkstream-peek/7281113#7281113 for the PeekableStream class

### DIFF
--- a/src/DotNetty.Handlers/Tls/PeekableStream.cs
+++ b/src/DotNetty.Handlers/Tls/PeekableStream.cs
@@ -1,0 +1,148 @@
+﻿// Credit http://stackoverflow.com/questions/2196767/c-implementing-networkstream-peek/7281113#7281113
+namespace DotNetty.Handlers.Tls
+{
+    using System;
+    using System.IO;
+
+    /// <summary>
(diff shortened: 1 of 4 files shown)


krish-gh:
Found some problem in e2e testing. Will reopen after fixing

nayato:
@krish-gh general comment on the approach in PR: idiomatic way of doing SNI would be to:
- write a new descendant of `ByteToMessageDecoder`
- override `Decode` in it, looking for Client Hello frame
- extract hostname from SNI extension in the frame
- find the certificate through lookup function user provided in a call to handler's constructor (e.g. `Func<string, Task<X509Certificate2>> certificateLookup`)
Netty's [`SniHandler`](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SniHandler.java#L112) does exactly that in a self-contained way (not involving any libraries). It makes more sense to do a port of `SniHandler` instead of complicating `TlsHandler`.

krish-gh:
Thanks. Will try this approach.

PR #219, opened 2017-03-25
SNI Support (ported netty SniHandler)

krish-gh:
Added support of SNI (https://tools.ietf.org/html/rfc3546#section-3.1) porting the netty variant.
N.B. The implementation had to take into consideration the way .net SslStream is used in existing TlsHandler in dotnetty. So the implementation is slightly different than netty which uses different ssl context based on the java variant.


nayato:
pls use this cert to demo SNI: https://1drv.ms/u/s!AmwDNZLGmm52oXSvSpcRI3pt7Yc4 Password is `password`

src/DotNetty.Handlers/Tls/SniHandler.cs
```diff
@@ -0,0 +1,260 @@
...
+        void Select(IChannelHandlerContext context, string hostName) => this.ReplaceHandler(context, hostName);
+
+        void ReplaceHandler(IChannelHandlerContext context, string hostName)
+        {
+            Contract.Requires(hostName != null);
+            var serverTlsSetting = new ServerTlsSettings(this.serverTlsSniSettings.ServerCertificateSelector(hostName), this.serverTlsSniSettings.NegotiateClientCertificate, this.serverTlsSniSettings.CheckCertificateRevocation, this.serverTlsSniSettings.EnabledProtocols);
+            var tlsHandler = new TlsHandler(this.sslStreamFactory, serverTlsSetting);
+            context.Channel.Pipeline.Replace(this, nameof(TlsHandler), tlsHandler);
+        }
+    }
```
nayato: why did you not port (or actually removed the port of) Read() that suppresses propagation?

src/DotNetty.Handlers/Tls/ServerTlsSniSettings.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace DotNetty.Handlers.Tls
+{
+    using System;
+    using System.Diagnostics.Contracts;
+    using System.Security.Authentication;
+    using System.Security.Cryptography.X509Certificates;
+
+    public sealed class ServerTlsSniSettings : TlsSettings
+    {
+        public ServerTlsSniSettings(Func<string, X509Certificate2> serverCertificateSelector)
```
nayato: >Func<string, X509Certificate2> [](start = 36, length = 30)

pls change to Func<string, Task<X509Certificate2>>. I can imagine pulling cert out of remote store / from disk lazily.

src/DotNetty.Handlers/Tls/SniHandler.cs
```diff
@@ -0,0 +1,260 @@
...
+                                                int serverNameLength = input.GetUnsignedShort(offset);
+                                                offset += 2;
+
+                                                if (serverNameLength <= 0 || extensionsLimit - offset < serverNameLength)
+                                                {
+                                                    continueLoop = false;
+                                                    break;
+                                                }
+
+                                                string hostname = input.ToString(offset, serverNameLength, Encoding.ASCII);
```
nayato: use Encoding.UTF8.GetString() to parse the string out of bytes (per RFC for TLS Extensions), then use IdnMapping.ToAscii()

src/DotNetty.Handlers/Tls/ServerTlsSniSettings.cs
krish-gh: Done

src/DotNetty.Handlers/Tls/SniHandler.cs
krish-gh: Done. But regarding to lower case conversion like netty for case insensitive hostname lookup, netcore ToLower method does not accept culture as input, so kept a todo there for net45, I passed en-US as culture like netty.

src/DotNetty.Handlers/Tls/SniHandler.cs
krish-gh: Okay the reason was suppress read is used in netty during the async sslcontext lookup being executed. In this case so far I was using sync cert lookup. But now since I changed the cert lookup async now as per your other comment, I have now ported the same read of netty.

src/DotNetty.Handlers/Tls/SniHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+                    TlsUtils.NotifyHandshakeFailure(context, e);
+                    throw e;
+                }
+            }
+        }
+
+        void Select(IChannelHandlerContext context, string hostName)
+        {
+            Contract.Requires(hostName != null);
+            var getCertTask = this.serverTlsSniSettings.ServerCertificateSelector(hostName);
```
nayato: pls switch to async / await. Yes, it will be an "async void" method, but you'll protect the body with try-catch and it'll be ok. It will reduce unnecessary complication of code + make sure continuation is dispatched back to executor.

krish-gh: @nayato Done. Which also means removed the suppress read logic and back to my previous implementation except the async cert lookup task.

src/DotNetty.Handlers/Tls/SniHandler.cs
```diff
@@ -0,0 +1,287 @@
...
+            catch (Exception ex)
+            {
+                this.ExceptionCaught(context, new DecoderException($"failed to get the Tls Certificate for {hostName}, {ex}"));
+            }
+        }
+
+        void ReplaceHandler(IChannelHandlerContext context, X509Certificate2 tlsCertificate)
+        {
+            Contract.Requires(tlsCertificate != null);
+            var serverTlsSetting = new ServerTlsSettings(tlsCertificate, this.serverTlsSniSettings.NegotiateClientCertificate, this.serverTlsSniSettings.CheckCertificateRevocation, this.serverTlsSniSettings.EnabledProtocols);
```
nayato: looking at the initialization pattern here and in tests it feels a bit too constrained to use the same settings for different endpoints. E.g. if you're co-hosting endpoints where one would negotiate client certs while the other wouldn't, or where you would allow older protocols on a separate endpoint. What do you think about changing the map to (string -> Task<ServerTlsSettings)?

test/DotNetty.Handlers.Tests/SniHandlerTest.cs
```diff
@@ -0,0 +1,296 @@
...
+                from isClient in boolToggle
+                from writeStrategyFactory in writeStrategyFactories
+                from protocol in protocols
+                from targetHost in CertMap.Keys
+                select new object[] { frameLengths, isClient, writeStrategyFactory(), protocol, targetHost };
+        }
+
+
+        [Theory]
+        [MemberData(nameof(GetTlsReadTestData))]
```
nayato: >GetTlsReadTestData [](start = 27, length = 18)

there is no need to run that extensive test suite (which exercises TlsHandler in fact).

nayato:
I think I'm through the review now. Basically reducing scope of testing for SNI to SNI itself is what's left to address.

test/DotNetty.Handlers.Tests/SniHandlerTest.cs
krish-gh: @nayato done. But haven't changed MemberData to InlineData since I am still feeding test data with multiple host names, so easier to do via MemberData. Just reduced the other inputs to 1 set in each.

krish-gh: @nayato CI build is failing with error "Error sending GitHubPullRequest notification: Error commenting on GitHub pull request: {"message":"Issue not viewable by nayato".....!

krish-gh:
#221

test/DotNetty.Handlers.Tests/SniHandlerTest.cs
```diff
@@ -0,0 +1,282 @@
...
+            }
+        }
+
+        public static IEnumerable<object[]> GetTlsWriteTestData()
+        {
+            var random = new Random(Environment.TickCount);
+            var lengthVariations =
+                new[]
+                {
+                    Enumerable.Repeat(0, 30).Select(_ => random.Next(0, 10) < 2 ? -1 : random.Next(0, 17000)).ToArray()
```
nayato: this generates 30 random data frames. pls replace with `new [] { 1 }`

test/DotNetty.Handlers.Tests/SniHandlerTest.cs
```diff
@@ -0,0 +1,282 @@
...
+        {
+        }
+
+        public static IEnumerable<object[]> GetTlsReadTestData()
+        {
+            var random = new Random(Environment.TickCount);
+            var lengthVariations =
+                new[]
+                {
+                    Enumerable.Repeat(0, 30).Select(_ => random.Next(0, 17000)).ToArray()
```
nayato: see comment on test data for write.

nayato:
@krish-gh that would be the last small thing to address. Thanks for pushing this through!

test/DotNetty.Handlers.Tests/SniHandlerTest.cs
krish-gh: @nayato done

test/DotNetty.Handlers.Tests/SniHandlerTest.cs
krish-gh: @nayato done

Issue #218, opened 2017-03-23
System.InvalidOperationException:“An asynchronous socket operation is already in progress using this SocketAsyncEventArgs instance.”


yyjdelete:
1. run `Discard.Server` at an remote PC(Never happens for localhost);
2. modify host in `appsettings.json` to the remote;
3. run `Discard.Client`;
4. wait about 5 to 20s(500 to 2000 times WriteAndFlushAsync);(may need use up tcp window, not sure)
5. See the Exception.

async Stack:
```
System.InvalidOperationException: An asynchronous socket operation is already in progress using this SocketAsyncEventArgs instance.
   at System.Net.Sockets.SocketAsyncEventArgs.StartConfiguring()
   at System.Net.Sockets.SocketAsyncEventArgs.set_BufferList(IList`1 value)
   at DotNetty.Transport.dll!DotNetty.Transport.Channels.Sockets.AbstractSocketChannel.PrepareWriteOperation(System.Collections.Generic.IList<System.ArraySegment<byte>> buffers):line 149
   at DotNetty.Transport.dll!DotNetty.Transport.Channels.Sockets.TcpSocketChannel.DoWrite(DotNetty.Transport.Channels.ChannelOutboundBuffer input):line 330
   at DotNetty.Transport.dll!DotNetty.Transport.Channels.AbstractChannel.AbstractUnsafe.Flush0():line 764
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Discard.Client.DiscardClientHandler.<GenerateTraffic>d__6.MoveNext() in F:\_lab\DotNetty\examples\Discard.Client\DiscardClientHandler.cs:line 43
```

FirstChanceException Stack:
```
System.InvalidOperationException: An asynchronous socket operation is already in progress using this SocketAsyncEventArgs instance.
   at System.Net.Sockets.SocketAsyncEventArgs.StartConfiguring()
   at System.Net.Sockets.SocketAsyncEventArgs.set_BufferList(IList`1 value)
[...]

PR #220, opened 2017-03-26
Fix possible concurrently flush(#218)

yyjdelete:


yyjdelete:
Add an ugly unit test. 😢

nayato:
test is rather complicated. I'll close this in favor of #250 where I've picked up your initial commit.

PR #231, opened 2017-04-06
Change type of RemoteCertificate property in TlsHandler to X509Certificate2

rajeevmv:


src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -64,19 +64,13 @@ public TlsHandler(Func<Stream, SslStream> sslStreamFactory, TlsSettings settings
 
         public X509Certificate LocalCertificate => this.sslStream.LocalCertificate;
```
nayato: can we do the same here?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -64,19 +64,13 @@ public TlsHandler(Func<Stream, SslStream> sslStreamFactory, TlsSettings settings
...
 
-        public X509Certificate RemoteCertificate => this.sslStream.RemoteCertificate;
+        // using workaround mentioned here: https://github.com/dotnet/corefx/issues/4510
+        public X509Certificate2 RemoteCertificate => new X509Certificate2(this.sslStream.RemoteCertificate?.Export(X509ContentType.Cert));
```
nayato: can you pls try safe casting to cert2 first and if that fails fallback to export?

src/DotNetty.Handlers/Tls/TlsHandler.cs
rajeevmv: done

src/DotNetty.Handlers/Tls/TlsHandler.cs
rajeevmv: done

PR #242, opened 2017-04-25
Porting HashedWheelTimer

nayato:
Motivation:
It is a common task to manage various timeouts in networking code. Normal high precision timers usually trade performance for precision which is a flexible requirement with timeouts.

Modifications:
- added HashedWheelTimer
- moved RejectedExecutionException to Common.Concurrency

Result:
- users may choose to use HWT in their code to avoid contentions due to frequent timer management.

### DIFF
--- a/src/DotNetty.Common/Utilities/HashedWheelTimer.cs
+++ b/src/DotNetty.Common/Utilities/HashedWheelTimer.cs
@@ -1,0 +1,705 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+
+#pragma warning disable 420
+
+namespace DotNetty.Common.Utilities
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Diagnostics.Contracts;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using DotNetty.Common.Concurrency;
+    using DotNetty.Common.Internal;
+    using DotNetty.Common.Internal.Logging;
+
+    public sealed class HashedWheelTimer : ITimer
+    {
+        static readonly IInternalLogger Logger =
+            InternalLoggerFactory.GetInstance<HashedWheelTimer>();
+
+        static int instanceCounter;
+        static int warnedTooManyInstances;
+
+        const int InstanceCountLimit = 64;
+
+        readonly Worker worker;
+        readonly XThread workerThread;
+        readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+
+        const int WorkerStateInit = 0;
+        const int WorkerStateStarted = 1;
+        const int WorkerStateShutdown = 2;
+        int workerStateVolatile = WorkerStateInit; // 0 - init, 1 - started, 2 - shut down
+
+        readonly long tickDuration;
+        readonly HashedWheelBucket[] wheel;
+        readonly int mask;
+        readonly CountdownEvent startTimeInitialized = new CountdownEvent(1);
+        readonly IQueue<HashedWheelTimeout> timeouts = PlatformDependent.NewMpscQueue<HashedWheelTimeout>();
+        readonly IQueue<HashedWheelTimeout> cancelledTimeouts = PlatformDependent.NewMpscQueue<HashedWheelTimeout>();
+        internal long PendingTimeouts;
+        readonly long maxPendingTimeouts;
+        long startTimeVolatile;
+
+        public HashedWheelTimer()
+            : this(TimeSpan.FromMilliseconds(100), 512, -1)
+        {
+        }
+
+        /// <summary>
+        /// Creates a new timer.
+        ///
+        /// </summary>
+        /// <param name="tickDuration">the duration between tick</param>
+        /// <param name="ticksPerWheel">the size of the wheel</param>
+        /// <param name=" maxPendingTimeouts">The maximum number of pending timeouts after which call to
+        /// <c>newTimeout</c> will result in <see cref="RejectedExecutionException"/> being thrown.
+        /// No maximum pending timeouts limit is assumed if this value is 0 or negative.</param>
+        /// <exception cref="ArgumentException">if either of <c>tickDuration</c> and <c>ticksPerWheel</c> is &lt;= 0</exception>
+        public HashedWheelTimer(
+            TimeSpan tickDuration,
+            int ticksPerWheel,
+            long maxPendingTimeouts)
+        {
+            if (tickDuration <= TimeSpan.Zero)
+            {
+                throw new ArgumentException("tickDuration must be greater than 0: " + tickDuration);
+            }
+            if (ticksPerWheel <= 0)
+            {
+                throw new ArgumentException("ticksPerWheel must be greater than 0: " + ticksPerWheel);
+            }
+
+            // Normalize ticksPerWheel to power of two and initialize the wheel.
+            this.wheel = CreateWheel(ticksPerWheel);
+            this.worker = new Worker(this);
+            this.mask = this.wheel.Length - 1;
+
+            // Convert tickDuration to nanos.
+            this.tickDuration = PreciseTimeSpan.FromTimeSpan(tickDuration).Ticks;
+
+            // Prevent overflow
+            if (this.tickDuration >= long.MaxValue / this.wheel.Length)
+            {
+                throw new ArgumentException(
+                    string.Format(
+                        "tickDuration: {0} (expected: 0 < tickDuration in nanos < {1}",
+                        tickDuration,
+                        long.MaxValue / this.wheel.Length));
+            }
+            this.workerThread = new XThread(st => this.worker.Run());
+
+            this.maxPendingTimeouts = maxPendingTimeouts;
+
+            if (Interlocked.Increment(ref instanceCounter) > InstanceCountLimit &&
+                Interlocked.CompareExchange(ref warnedTooManyInstances, 1, 0) == 0)
+            {
+                ReportTooManyInstances();
+            }
+        }
+
+        ~HashedWheelTimer()
+        {
+            // This object is going to be GCed and it is assumed the ship has sailed to do a proper shutdown. If
+            // we have not yet shutdown then we want to make sure we decrement the active instance count.
+            if (Interlocked.Exchange(ref this.workerStateVolatile, WorkerStateShutdown) != WorkerStateShutdown)
+            {
+                Interlocked.Decrement(ref instanceCounter);
+            }
+        }
+
+        internal CancellationToken CancellationToken => this.cancellationTokenSource.Token;
+
+        PreciseTimeSpan StartTime
+        {
+            get { return PreciseTimeSpan.FromTicks(Volatile.Read(ref this.startTimeVolatile)); }
+            set { Volatile.Write(ref this.startTimeVolatile, value.Ticks); }
+        }
+
+        static HashedWheelBucket[] CreateWheel(int ticksPerWheel)
+        {
+            if (ticksPerWheel <= 0)
+            {
+                throw new ArgumentException(
+                    "ticksPerWheel must be greater than 0: " + ticksPerWheel);
+            }
+            if (ticksPerWheel > 1073741824)
+            {
+                throw new ArgumentException(
+                    "ticksPerWheel may not be greater than 2^30: " + ticksPerWheel);
+            }
+
+            ticksPerWheel = NormalizeTicksPerWheel(ticksPerWheel);
+            var wheel = new HashedWheelBucket[ticksPerWheel];
+            for (int i = 0; i < wheel.Length; i++)
+            {
+                wheel[i] = new HashedWheelBucket();
+            }
+            return wheel;
+        }
+
+        static int NormalizeTicksPerWheel(int ticksPerWheel)
+        {
+            int normalizedTicksPerWheel = 1;
+            while (normalizedTicksPerWheel < ticksPerWheel)
+            {
+                normalizedTicksPerWheel <<= 1;
+            }
+            return normalizedTicksPerWheel;
+        }
+
+        /// <summary>
+        /// Starts the background thread explicitly. The background thread will
+        /// start automatically on demand even if you did not call this method.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">if this timer has been
+        /// stopped already.</exception>
+        public void Start()
+        {
+            switch (Volatile.Read(ref this.workerStateVolatile))
+            {
+                case WorkerStateInit:
+                    if (Interlocked.CompareExchange(ref this.workerStateVolatile, WorkerStateStarted, WorkerStateInit) == WorkerStateInit)
+                    {
+                        this.workerThread.Start();
+                    }
+                    break;
+                case WorkerStateStarted:
+                    break;
+                case WorkerStateShutdown:
+                    throw new InvalidOperationException("cannot be started once stopped");
+                default:
+                    throw new InvalidOperationException("Invalid WorkerState");
+            }
+
+            // Wait until the startTime is initialized by the worker.
+            while (this.StartTime == PreciseTimeSpan.Zero)
+            {
+                this.startTimeInitialized.Wait(this.CancellationToken);
+            }
+        }
+
+        public ISet<ITimeout> Stop()
+        {
+            if (XThread.CurrentThread == this.workerThread)
+            {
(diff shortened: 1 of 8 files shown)


src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                        break;
+                    }
+                    if (timeout.State == HashedWheelTimeout.StCancelled)
+                    {
+                        // Was cancelled in the meantime.
+                        continue;
+                    }
+
+                    long calculated = (timeout.Deadline.Ticks + this.owner.tickDuration - 1) / this.owner.tickDuration; // ceiling to timeout later rather than earlier
+                    timeout.RemainingRounds = (calculated - this.tick) / this.owner.wheel.Length;
```
mastermanu: So this can be negative then, right?

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                    int stopIndex = (int)(ticks & this.owner.mask);
+
+                    HashedWheelBucket bucket = this.owner.wheel[stopIndex];
+                    bucket.AddTimeout(timeout);
+                }
+            }
+
+            void ProcessCancelledTasks()
+            {
+                while (true)
```
mastermanu: No limit here as well? Or I guess CancelledTasks always take priority since they drain the queue as quickly as possible?

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                long deadline = this.owner.tickDuration * (this.tick + 1);
+
+                while (true)
+                {
+                    TimeSpan currentTime = (PreciseTimeSpan.FromStart - this.owner.StartTime).ToTimeSpan();
+                    TimeSpan sleepTime = CeilTimeSpanToMilliseconds(TimeSpan.FromTicks(deadline - currentTime.Ticks));
+
+                    if (sleepTime <= TimeSpan.Zero)
+                    {
+                        if (currentTime.Ticks == long.MinValue)
```
mastermanu: if this is happening consistently, then does that indicate that there are too many timers in the system to keep up with the tick count?

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                    try
+                    {
+                        long sleepTimeMs = sleepTime.Ticks / TimeSpan.TicksPerMillisecond; // we've already rounded so no worries about the remainder > 0 here
+                        Contract.Assert(sleepTimeMs <= int.MaxValue);
+                        delay = Task.Delay((int)sleepTimeMs, this.owner.CancellationToken);
+                        delay.Wait();
+                    }
+                    catch (AggregateException) when (delay != null && delay.IsCanceled)
+                    {
+                        if (Volatile.Read(ref this.owner.workerStateVolatile) == WorkerStateShutdown)
```
mastermanu: let's say this if condition isn't true because of some bug? what should we do then?

I guess it will just keep spinning until sleepTime <= 0 and then bail?

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                        if (!this.owner.timeouts.TryDequeue(out timeout))
+                        {
+                            break;
+                        }
+                        if (!timeout.Cancelled)
+                        {
+                            this.unprocessedTimeouts.Add(timeout);
+                        }
+                    }
+                    this.ProcessCancelledTasks();
```
mastermanu: what's the reason for this? Just for GC purposes?

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                // So this means that we will have a GC latency of max. 1 tick duration which is good enough. This way
+                // we can make again use of our MpscLinkedQueue and so minimize the locking / overhead as much as possible.
+                this.timer.cancelledTimeouts.TryEnqueue(this);
+                return true;
+            }
+
+            internal void Remove()
+            {
+                HashedWheelBucket bucket = this.Bucket;
+                if (bucket != null)
```
mastermanu: when is Bucket null?

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+            internal void Expire()
+            {
+                if (!this.CompareAndSetState(StInit, StExpired))
+                {
+                    return;
+                }
+
+                try
+                {
+                    this.Task.Run(this);
```
mastermanu: if Run() blocks, then I guess we gum up the whole timer system, right?

mastermanu: should we assert bucket = null by this point?

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                            throw new InvalidOperationException(
+                                string.Format(
+                                    "timeout.deadline (%d) > deadline (%d)",
+                                    timeout.Deadline,
+                                    deadline));
+                        }
+                    }
+                    else if (timeout.Cancelled)
+                    {
+                        next = this.Remove(timeout);
```
mastermanu: if deadlines start to really fall behind by orders of seconds, are we tracing some sort of warning / error so we know we are backed up? or is it not necessary?

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                this.timer = timer;
+                this.Task = task;
+                this.Deadline = deadline;
+            }
+
+            public ITimer Timer => this.timer;
+
+            public ITimerTask Task { get; }
+
+            public bool Cancel()
```
mastermanu: if the HashedWheelTimer is shutdown and then later someone calls Cancel(), then this will just leak an entry on the IQueue object on the timer class, right? should we not enqueue if the timer is shutdown?

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                        {
+                            // The timeout was placed into a wrong slot. This should never happen.
+                            throw new InvalidOperationException(
+                                string.Format(
+                                    "timeout.deadline (%d) > deadline (%d)",
+                                    timeout.Deadline,
+                                    deadline));
+                        }
+                    }
+                    else if (timeout.Cancelled)
```
mastermanu: don't do this from two places. you already have a place where you are reading off of the canceled queue and removing it at that point. otherwise Remove() is called twice for the same timeout object potentially and from what I can tell it can lead to weird behavior (such as double decrementing pending timer count)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                    this.tail.Next = timeout;
+                    timeout.Prev = this.tail;
+                    this.tail = timeout;
+                }
+            }
+
+            /// <summary>
+            /// Expire all <see cref="HashedWheelTimeout"/>s for the given <c>deadline</c>.
+            /// </summary>
+            public void ExpireTimeouts(TimeSpan deadline)
```
mastermanu: we expect this to be efficient because we expect most timers to be canceled before it comes back to the bucket again. would be good to get a trace if the assumption we made is incorrect (in which case, user would tune this thing to have more buckets or a greater tick duration potentially)

src/DotNetty.Common/Utilities/ITimeout.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// Returns {@code true} if and only if the <see cref="ITimerTask"/> associated
+        /// with this handle has been expired.
+        /// </summary>
+        bool Expired { get; }
+
+        /// <summary>
+        /// Returns <c>true</c> if and only if the <see cref="ITimerTask"/> associated
+        /// with this handle has been cancelled.
+        /// </summary>
+        bool Cancelled { get; }
```
nayato: Porting down to the last typo :)

---
In reply to: [114472566](https://github.com/Azure/DotNetty/pull/242#discussion_r114472566) [](ancestors = 114472566)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+                case WorkerStateStarted:
+                    break;
+                case WorkerStateShutdown:
+                    throw new InvalidOperationException("cannot be started once stopped");
+                default:
+                    throw new InvalidOperationException("Invalid WorkerState");
+            }
+
+            // Wait until the startTime is initialized by the worker.
+            while (this.StartTime == PreciseTimeSpan.Zero)
```
nayato: makes no sense on a second look

---
In reply to: [114474586](https://github.com/Azure/DotNetty/pull/242#discussion_r114474586) [](ancestors = 114474586)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
nayato: I'll add a comment

---
In reply to: [114477290](https://github.com/Azure/DotNetty/pull/242#discussion_r114477290) [](ancestors = 114477290)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
nayato: +1

---
In reply to: [114478174](https://github.com/Azure/DotNetty/pull/242#discussion_r114478174) [](ancestors = 114478174)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
nayato: +!

---
In reply to: [114479707](https://github.com/Azure/DotNetty/pull/242#discussion_r114479707) [](ancestors = 114479707)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+        public HashedWheelTimer(
+            TimeSpan tickDuration,
+            int ticksPerWheel,
+            long maxPendingTimeouts)
+        {
+            if (tickDuration <= TimeSpan.Zero)
+            {
+                throw new ArgumentException("tickDuration must be greater than 0: " + tickDuration);
+            }
+            if (Math.Ceiling(tickDuration.TotalMilliseconds) > int.MaxValue)
```
nayato: this is basic protection for correctness. We'll never need more than 1 sec I imagine.

---
In reply to: [114473436](https://github.com/Azure/DotNetty/pull/242#discussion_r114473436) [](ancestors = 114473436)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+        /// </summary>
+        /// <param name="tickDuration">the duration between tick</param>
+        /// <param name="ticksPerWheel">the size of the wheel</param>
+        /// <param name="maxPendingTimeouts">The maximum number of pending timeouts after which call to
+        /// <c>newTimeout</c> will result in <see cref="RejectedExecutionException"/> being thrown.
+        /// No maximum pending timeouts limit is assumed if this value is 0 or negative.</param>
+        /// <exception cref="ArgumentException">if either of <c>tickDuration</c> and <c>ticksPerWheel</c> is &lt;= 0</exception>
+        public HashedWheelTimer(
+            TimeSpan tickDuration,
+            int ticksPerWheel,
```
nayato: that would be inconvenient in calculations (due to need for conversion)

---
In reply to: [114473422](https://github.com/Azure/DotNetty/pull/242#discussion_r114473422) [](ancestors = 114473422)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+            // Prevent overflow
+            if (this.tickDuration >= long.MaxValue / this.wheel.Length)
+            {
+                throw new ArgumentException(
+                    string.Format(
+                        "tickDuration: {0} (expected: 0 < tickDuration in nanos < {1}",
+                        tickDuration,
+                        long.MaxValue / this.wheel.Length));
+            }
+            this.workerThread = new XThread(st => this.worker.Run());
```
nayato: yes, thread won't start until later use of the HWT object itself which cannot happen before the ctor is completed.

---
In reply to: [114473937](https://github.com/Azure/DotNetty/pull/242#discussion_r114473937) [](ancestors = 114473937)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+
+                    // Notify the other threads waiting for the initialization at start().
+                    this.owner.startTimeInitialized.Signal();
+
+                    do
+                    {
+                        TimeSpan deadline = this.WaitForNextTick();
+                        if (deadline > TimeSpan.Zero)
+                        {
+                            int idx = (int)(this.tick & this.owner.mask);
```
nayato: yes

---
In reply to: [114475721](https://github.com/Azure/DotNetty/pull/242#discussion_r114475721) [](ancestors = 114475721)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
nayato: yes and it's supported

---
In reply to: [114476429](https://github.com/Azure/DotNetty/pull/242#discussion_r114476429) [](ancestors = 114476429)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
nayato: cleaning up canceled is a priority as it is the least amount of work to process scheduled timeouts.

---
In reply to: [114476619](https://github.com/Azure/DotNetty/pull/242#discussion_r114476619) [](ancestors = 114476619)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
nayato: IMO hypothetical bug chasing is not productive. and yes, it would.

---
In reply to: [114476824](https://github.com/Azure/DotNetty/pull/242#discussion_r114476824) [](ancestors = 114476824)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
nayato: correct; you can choose in your ITimeout implementation to schedule callback on a threadpool to avoid such issues. No limitations from HWT itself though.

---
In reply to: [114477953](https://github.com/Azure/DotNetty/pull/242#discussion_r114477953) [](ancestors = 114477953)

nayato: why? if it expires it means we drove it to this state.

---
In reply to: [114478093](https://github.com/Azure/DotNetty/pull/242#discussion_r114478093) [](ancestors = 114478093)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
```diff
@@ -0,0 +1,712 @@
...
+            this.timeouts.TryEnqueue(timeout);
+            return timeout;
+        }
+
+        bool ShouldLimitTimeouts() => this.maxPendingTimeouts > 0;
+
+        static void ReportTooManyInstances() =>
+            Logger.Error($"You are creating too many {nameof(HashedWheelTimer)} instances. {nameof(HashedWheelTimer)} is a shared resource that must be reused across the process,so that only a few instances are created.");
+
+        static TimeSpan CeilTimeSpanToMilliseconds(TimeSpan time)
```
nayato: yes

---
In reply to: [114475437](https://github.com/Azure/DotNetty/pull/242#discussion_r114475437) [](ancestors = 114475437)

src/DotNetty.Common/Utilities/HashedWheelTimer.cs
nayato: this is preventable by using a max on timer limit. Although you will get exceptions on over-scheduling then..

---
In reply to: [114476742](https://github.com/Azure/DotNetty/pull/242#discussion_r114476742) [](ancestors = 114476742)

PR #247, opened 2017-05-04
Make ReadOnlyByteBufferStream seekable

damonbarry:
The [Protocol Gateway](https://github.com/Azure/azure-iot-protocol-gateway) uses ReadOnlyByteBufferStream to create an IotHubClientMessage. The IoTHub device SDK then calls .Length on the underlying stream, which is a problem because ReadOnlyByteBufferStream isn't currently seekable. The device SDK will be updated to call CanSeek() first, but it would also be good to add seekable support to ReadOnlyByteBufferStream.


damonbarry:
Missed some cases, which other tests exposed. I'll push more commits...

src/DotNetty.Buffers/ReadOnlyByteBufferStream.cs
```diff
@@ -29,7 +29,7 @@ public ReadOnlyByteBufferStream(IByteBuffer buffer, bool releaseReferenceOnClosu
 
         public override long Length
         {
-            get { throw new NotSupportedException(); }
+            get => this.buffer.ReadableBytes;
```
yyjdelete: IByteBuffer.ReadableBytes may not equals to Stream.Length since it decrease after read.

damonbarry: Ah, of course. The most recent changes I made were to handle/expect the decrease in ReadableBytes, but I didn't put two and two together...

I'll rework this. Thanks!

damonbarry: Fixed. I pushed (rebased) commits to take care of this.

src/DotNetty.Buffers/ReadOnlyByteBufferStream.cs
```diff
@@ -23,24 +23,26 @@ public ReadOnlyByteBufferStream(IByteBuffer buffer, bool releaseReferenceOnClosu
...
 
-        public override bool CanSeek => false;
+        public override bool CanSeek => true;
 
         public override bool CanWrite => false;
 
         public override long Length
         {
-            get { throw new NotSupportedException(); }
+            get => this.buffer.ReaderIndex + this.buffer.ReadableBytes;
```
nayato: by definition it is this.buffer.WriterIndex

src/DotNetty.Buffers/ReadOnlyByteBufferStream.cs
```diff
@@ -23,24 +23,26 @@ public ReadOnlyByteBufferStream(IByteBuffer buffer, bool releaseReferenceOnClosu
...
+            get => this.buffer.ReaderIndex;
+            set => this.buffer.SetReaderIndex((int)value);
         }
 
         public override int Read(byte[] output, int offset, int count)
         {
-            int read = Math.Min(count - offset, this.buffer.ReadableBytes);
+            if (offset + count > output.Length) throw new ArgumentException();
+            int sourceBytes = Math.Max(0, this.buffer.ReadableBytes - this.buffer.ReaderIndex);
+            int read = Math.Min(count, sourceBytes);
```
nayato: >sourceBytes [](start = 39, length = 11)

shouldn't you just use this.buffer.ReadableBytes instead of sourceBytes?

src/DotNetty.Buffers/ReadOnlyByteBufferStream.cs
```diff
@@ -67,7 +69,17 @@ protected override void Dispose(bool disposing)
...
         public override long Seek(long offset, SeekOrigin origin)
         {
-            throw new NotSupportedException();
+            if (origin == SeekOrigin.Current)
+            {
+                offset += this.Position;
+            }
+            else if (origin == SeekOrigin.End)
+            {
+                offset += this.buffer.Capacity;
```
nayato: should be `offset += this.buffer.WriterIndex;` as we consider WriterIndex the end of readable space

src/DotNetty.Buffers/ReadOnlyByteBufferStream.cs
damonbarry: Yep you're right, thanks for catching this. I missed a unit test that would've caught this dumb error; adding it...

src/DotNetty.Buffers/ReadOnlyByteBufferStream.cs
damonbarry: I remember you telling me this, but I didn't really see it in the code. Regardless, this is simpler/cleaner. I'll do it.

src/DotNetty.Buffers/ReadOnlyByteBufferStream.cs
damonbarry: Yep. I'll update the code...

damonbarry:
@nayato My latest push should address your comments. Thanks!

src/DotNetty.Buffers/ReadOnlyByteBufferStream.cs
```diff
@@ -23,24 +23,22 @@ public ReadOnlyByteBufferStream(IByteBuffer buffer, bool releaseReferenceOnClosu
...
-            get { throw new NotSupportedException(); }
-            set { throw new NotSupportedException(); }
+            get => this.buffer.ReaderIndex;
+            set => this.buffer.SetReaderIndex((int)value);
         }
 
         public override int Read(byte[] output, int offset, int count)
         {
-            int read = Math.Min(count - offset, this.buffer.ReadableBytes);
+            if (offset + count > output.Length) throw new ArgumentException();
```
nayato: per coding style pls show code block explicitly (curly braces). Also, pls add a msg like `$"The sum of {nameof(offset)} and {nameof(count)} is larger than the {nameof(output)} length"`.

damonbarry: Thanks, will do!

test/DotNetty.Buffers.Tests/TestByteBuffer.cs
```diff
@@ -0,0 +1,600 @@
...
+{
+    using System;
+    using System.IO;
+    using System.Linq;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using DotNetty.Common;
+    using DotNetty.Common.Utilities;
+    class TestByteBuffer : IByteBuffer
```
nayato: this can be done with much less code using Moq :)

damonbarry:
@nayato I'm new to Moq, but take a look at 53faca3...is it better or worse than what I had?

nayato:
Looks good. Merging. Thanks!

Issue #36, opened 2015-10-12
Port Buffer.UnpooledTests


Aaronontheweb:
https://github.com/netty/netty/blob/master/buffer/src/test/java/io/netty/buffer/UnpooledTest.java - need to verify that the current `Unpooled` class works as intended

StormHub:
This has already been ported by #287 
https://github.com/netty/netty/blob/4.1/buffer/src/test/java/io/netty/buffer/UnpooledTest.java
I think we can close this.

PR #251, opened 2017-05-13
Expanded Unpooled functionality. Ported UnpooledTest(#36)

ScarletKuro:


src/DotNetty.Buffers/Unpooled.cs
```diff
@@ -58,6 +85,63 @@ public static IByteBuffer WrappedBuffer(byte[] array, int offset, int length)
...
+                    break;
+                }
+
+                if (a.Length > 0)
+                {
+                    components.Add(WrappedBuffer(a));
+                }
+            }
+
+            return components.Any() ? new CompositeByteBuffer(Allocator, maxNumComponents, components) : Empty;
```
nayato: `.Any()` -> `.Count > 0`

nayato:
looks good except for the comments above.

src/DotNetty.Common/Utilities/ArrayExtensions.cs
```diff
@@ -72,5 +72,44 @@ public static void Fill<T>(this T[] array, int offset, int count, T value)
...
             }
         }
+
+        /// <summary>
+        ///     Merge the byte arrays into one byte array.
+        /// </summary>
+        public static byte[] CombineBytes(this byte[][] arrays)
+        {
+            int newlength = 0;
+            arrays.ForEach(delegate (byte[] array) { newlength += array.Length; });
```
nayato: how about doing smth like:
```
public long SumLengths<T>(T arrays)
  where T: IEnumerable<Array>
{
  int sum = 0;
  foreach (var array in arrays)
  {
    sum += array.Length;
  }
  return sum;
}```

ScarletKuro: Hm. Sure. Add it as extension? And imma replace the ```int sum``` to ```long sum```.
Actually, not sure what to do with IByteBuffer then, since you need to count the ReadableBytes, or keep the ForEach anyway or just put another foreach above in ```IByteBuffer CopiedBuffer(params IByteBuffer[] buffers)``` to count the length

nayato: TBH, looking at actual three lines of ForEach body, pls just use `foreach (var item in array)` in place and don't add any extension methods for such a trivial thing.

ScarletKuro:
Done, i will leave the CombineBytes as extension tho, since i think merging bytes is useful to have?
Do i need to rebase and squash?

PR #260, opened 2017-07-01
Always use uppercase for HexDump

yyjdelete:
The current version get `0a` for `0x0A`, but `1A` for `0x1A` for value, and always upper for Prefix. ~The patch make it always output lowercase as netty does.~
***Which one should be used? Upper or Lower?***


yyjdelete:
https://github.com/Azure/DotNetty/blob/dev/src/DotNetty.Buffers/ByteBufferUtil.cs#L572
Since `RowPrefix` is already uppercase, should all things be changed to uppercase instead?

nayato:
@yyjdelete uppercase is much more common for hex representations with exception of hash strings. For byte buffer dumps I'd vote for upper case.

PR #263, opened 2017-07-13
TlsHandler drives handshake to completion with AutoRead=false

nayato:


nayato:
@alextolp please take a look

PR #270, opened 2017-07-30
Libuv based tcp transport.

StormHub:


src/DotNetty.Transport.Libuv/Native/SpscLinkedQueue.cs
```diff
@@ -0,0 +1,203 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace DotNetty.Transport.Libuv.Native
+{
+    using System.Diagnostics.Contracts;
+    using System.Threading;
+
+    sealed class SpscLinkedQueue<T> : BaseLinkedQueue<T>
```
nayato: >SpscLinkedQueue [](start = 17, length = 15)

should we move it to Common?

src/DotNetty.Transport.Libuv/LoopExecutor.cs
```diff
@@ -0,0 +1,272 @@
...
+            }
+            catch (Exception exception)
+            {
+                Logger.Error("{}: stop loop failed", exception);
+            }
+        }
+
+        protected virtual void Shutdown() => this.CancelScheduledTasks();
+
+        void RunAllTasks()
```
nayato: do we not need to introduce a break out like here?

test/DotNetty.Transport.Tests.Performance/Libuv/TcpPerfSpecs.cs
```diff
@@ -0,0 +1,144 @@
...
+            NumberOfIterations = IterationCount, RunMode = RunMode.Iterations)]
+        [CounterMeasurement(InboundThroughputCounterName)]
+        [CounterMeasurement(OutboundThroughputCounterName)]
+        [GcMeasurement(GcMetric.TotalCollections, GcGeneration.AllGc)]
+        [MemoryMeasurement(MemoryMetric.TotalBytesAllocated)]
+        public void TcpDuplexThroughput10MessagesPerFlush(BenchmarkContext context)
+        {
+            this.clientChannel.EventLoop.Execute(() =>
+            {
+                for (int i = 0; i < WriteCount; i++)
```
nayato: can we run from within the loop here? E.g.
[https://github.com/Azure/DotNetty/blob/dev/test/DotNetty.Transport.Tests.Performance/Sockets/TcpSocketChannelPerfSpec.cs#L135](https://github.com/Azure/DotNetty/blob/dev/test/DotNetty.Transport.Tests.Performance/Sockets/TcpSocketChannelPerfSpec.cs#L135)
it avoids cross-thread work posting which is not what we want reflected here.

PR #272, opened 2017-08-01
Fixes #271, stops sharing ClosedChannelException

### LINKED ISSUE #271: TcpServerSocketChannel will shutdown if AcceptAsync throws synchronously
We need to add retries if AcceptAsync throws synchronously (unless the Socket was actually closed intentionally). We can use the SocketError enum on a SocketException to figure out whether or not to retry

nayato:


src/DotNetty.Transport/Channels/Sockets/TcpServerSocketChannel.cs
```diff
@@ -179,24 +203,29 @@ public override void FinishRead(SocketChannelAsyncOperation operation)
...
+                        exception = ex;
+                    }
                     catch (Exception ex)
                     {
-                        var asSocketException = ex as SocketException;
-                        if (asSocketException == null || asSocketException.SocketErrorCode != SocketError.WouldBlock)
+                        Logger.Warn("Failed to create a new channel from an accepted socket.", ex);
+                        exception = ex;
+
+                        if (connectedSocket != null)
```
mastermanu: let's always dispose connectedSocket for socket exceptions if its not null.

basically, we can just remove the catch (SocketException ex) block

PR #285, opened 2017-08-26
Realigning API of IdleStateHandler, and some bugfix

yyjdelete:
* Align API and Test of IdleStateHandler with netty 4.1.15([f9003293...7460d90a](https://github.com/netty/netty/commits/4.1/handler/src/main/java/io/netty/handler/timeout))
* Fix EmbeddedChannel.ReleaseAll crashes when queue is null
* Fix ReadTimeoutHandler don't close channel when timeout
* Change IdleStateHandler.ChannelIdle and WriteTimeoutHandler.WriteTimedOut to virtual


nayato:
thanks!

PR #287, opened 2017-09-02
Re align with Netty buffers.

StormHub:


StormHub:
Deprecated ByteOrder and added LE methods into byte buffer separately
Exposing arena metrics
Greatly increased unit test coverage for all buffer types

StormHub:
Changed microbench to use BenchmarkDotNet

nayato:
That's a great work pushing this. Thanks a lot!

PR #289, opened 2017-09-19
Updating the Protobuf lib

wackoisgod:
Updating to the latest protobuf


nayato:
Why is this necessary? I.e. by increasing the version we're simply raising the bar to comply with dotnetty's dependency requirements. If there was no security concerns addressed since the version we have or explicit pledge to upgrade, I'd rather keep the current version. User can always choose to update the dependency in their project.

PR #293, opened 2017-10-09
Change name of prop `numHeapArenas` to match others and logging output

yyjdelete:
`dotNetty.allocator.numHeapArenas` -> `io.netty.allocator.numHeapArenas`


nayato:
lgtm

PR #294, opened 2017-10-09
Make it able to extends some FrameDecoder

yyjdelete:


yyjdelete:
Or change all `protected internal override void Decode(IChannelHandlerContext ctx, IByteBuffer input, List<object> output)` to `sealed`?

nayato:
lgtm

PR #297, opened 2017-10-26
Make ReadListeningHandler thread-safe

alextolp:


nayato:
Thanks for pulling this through

PR #299, opened 2017-10-28
Linux socket send bugfix

caozhiyuan:
fix .net core linux socket bug, use sendasync and fetch max 1024 outbondbuffer entry in linux
.net core issue
https://github.com/dotnet/corefx/issues/23416
https://github.com/dotnet/corefx/issues/23672

### DIFF
--- a/src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
+++ b/src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
@@ -313,7 +313,7 @@ namespace DotNetty.Transport.Channels
         // Clear all ByteBuffer from the array so these can be GC'ed.
         // See https://github.com/netty/netty/issues/3837
         void ClearNioBuffers() => NioBuffers.Value.Clear();
-        
+
         ///
         ///Returns an array of direct NIO buffers if the currently pending messages are made of {@link ByteBuf} only.
         ///{@link #nioBufferCount()} and {@link #nioBufferSize()} will return the number of NIO buffers in the returned
@@ -323,15 +323,20 @@ namespace DotNetty.Transport.Channels
         ///{@link AbstractChannel#doWrite(ChannelOutboundBuffer)}.
         ///Refer to {@link NioSocketChannel#doWrite(ChannelOutboundBuffer)} for an example.
         ///</p>
-        ///
-        public List<ArraySegment<byte>> GetSharedBufferList()
+        /// @param maxCount The maximum amount of buffers that will be added to the return value.
+        /// @param maxBytes A hint toward the maximum number of bytes to include as part of the return value. Note that this
+        ///                 value maybe exceeded because we make a best effort to include at least 1 {@link ByteBuffer}
+        ///                 in the return value to ensure write progress is made.
+        /// 
+        public List<ArraySegment<byte>> GetSharedBufferList(int maxCount, long maxBytes = int.MaxValue)
         {
             long nioBufferSize = 0;
             InternalThreadLocalMap threadLocalMap = InternalThreadLocalMap.Get();
             List<ArraySegment<byte>> nioBuffers = NioBuffers.Get(threadLocalMap);
             Entry entry = this.flushedEntry;
-            int i = 0;
-            while (this.IsFlushedEntry(entry) && entry.Message is IByteBuffer) {
+            int nioBufferCount = 0;
+            while (this.IsFlushedEntry(entry) && entry.Message is IByteBuffer)
+            {
                 if (!entry.Cancelled)
                 {
                     var buf = (IByteBuffer)entry.Message;
@@ -340,7 +345,7 @@ namespace DotNetty.Transport.Channels
 
                     if (readableBytes > 0)
                     {
-                        if (int.MaxValue - readableBytes < nioBufferSize)
+                        if (maxBytes - readableBytes < nioBufferSize && nioBufferCount != 0)
                         {
                             // If the nioBufferSize + readableBytes will overflow an Integer we stop populate the
                             // ByteBuffer array. This is done as bsd/osx don't allow to write more bytes then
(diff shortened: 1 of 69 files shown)


caozhiyuan:
in my dev branch has rpc client and server examples

caozhiyuan:
@nayato  can review it and merge?

src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
```diff
@@ -359,6 +361,12 @@ public List<ArraySegment<byte>> GetSharedBufferList()
                             //noinspection ConstantValueVariableUse
                             entry.Count = count = buf.IoBufferCount;
                         }
+                        i += count;
+                        if (i > 1024 && Util.IsLinux)
```
StormHub: Having a platform specific check and some hard coded limit here does not look like a very good idea to me. I think it might make more sense to introduce some channel config option like maxBytesPerGatheringWrite so that the outbound buffer behaves consistently across all platforms. Similar to netty Epoll config
https://github.com/netty/netty/blob/b215794de31f28355e4469fcc04782f55076c80c/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollChannelConfig.java#L33

caozhiyuan: it fix in .net core sdk 2.0.3, it can be removed. also you can review this code and change it better. now socket send will not send full buffer in linux when retuen sockererror.success. ,it dotnetty need fix.

StormHub: First of all, the outbound buffer cached list already has the limit of 1024
https://github.com/StormHub/DotNetty/blob/dev/src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs#L704
In original Netty, the buffer array is filled with 1024 slots per batch.
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java#L395
There are really two limits here
1. Maximum bytes gathered
2. Number of io buffer slots
If you do the same above, there is no need for some platform specific checks.

caozhiyuan: 1、you can debug it,i know that, java has check iovmax in socket channle . 
 2、it just decalre list capacity ， when you add item more capacity will auto change in .net

caozhiyuan: i will change my code like java ,thanks.

StormHub: For minimum change, what you have to do is to have a limit passed in similar to what Netty does
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java#L389
And just pass the limit in from socket channel
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/socket/nio/NioSocketChannel.java#L387

It is much better than some platform specific checks hiding in the outbound buffer.

caozhiyuan:
@StormHub  i change code to minimum change , code run ok on dotnet core sdk 2.0.3 or more. it has bug in dotnet core sdk 2.0.0

HenrikPoulsen:
It doesn't look like it. Looking at the issues linked in the first post the fix seems to have been pushed to 2.1.0. And after updating to 2.0.4 and running my repro project I wrote about in #323 it is still reproducible. So it does seem necessary to me.

StormHub:
@nayato there are still lots of ppl using netstandard 1.3 and the migration takes some time. Also, having a buffer count limit align with Netty nothing platform specific.

src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
```diff
@@ -6,6 +6,7 @@ namespace DotNetty.Transport.Channels
     using System;
     using System.Collections.Generic;
     using System.Diagnostics.Contracts;
+    using System.Runtime.InteropServices;
```
nayato: is it used?

src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
```diff
@@ -236,10 +237,10 @@ bool Remove0(Exception cause, bool notifyWritability)
                 ReferenceCountUtil.SafeRelease(msg);
 
                 Util.SafeSetFailure(promise, cause, Logger);
-                if (promise != TaskCompletionSource.Void && !promise.TrySetException(cause))
-                {
-                    Logger.Warn($"Failed to mark a promise as failure because it's done already: {promise}", cause);
-                }
+                //if (promise != TaskCompletionSource.Void && !promise.TrySetException(cause))
```
nayato: remove altogether

src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
```diff
@@ -379,11 +387,25 @@ public List<ArraySegment<byte>> GetSharedBufferList()
                                 // of Object allocation
                                 entry.Buffers = nioBufs = buf.GetIoBuffers();
                             }
-                            foreach (ArraySegment<byte> b in nioBufs)
+                            for (int i = 0; i < nioBufs.Length && nioBufferCount < maxCount; ++i)
```
nayato: `i++` please. I suppose it won't change the meaning but does confuse for a second the way it is right now

src/DotNetty.Transport/Channels/ChannelOutboundBuffer.cs
```diff
@@ -607,10 +644,10 @@ internal void Close(ClosedChannelException cause)
                     {
                         ReferenceCountUtil.SafeRelease(e.Message);
                         Util.SafeSetFailure(e.Promise, cause, Logger);
-                        if (e.Promise != TaskCompletionSource.Void && !e.Promise.TrySetException(cause))
-                        {
-                            Logger.Warn($"Failed to mark a promise as failure because it's done already: {e.Promise}", cause);
-                        }
+                        //if (e.Promise != TaskCompletionSource.Void && !e.Promise.TrySetException(cause))
```
nayato: remove instead

caozhiyuan: it's ok ， in linux  socket send would not send all buffer ， so  after send bufferlist it need get left not sent buffer

Issue #308, opened 2017-11-13
ResourceLeakDetector sometimes report current using buffer as leak in 0.4.7


yyjdelete:
Seems it works well with 0.4.6

@StormHub
A lot of things changed after 0.4.6, so don't what make the difference.
Is an extra `GC.SuppressFinalize(GCNotice)` is needed? (But not see in 0.4.6 either)

<details>
<summary>Test Code</summary>

```cs

using DotNetty.Buffers;
using DotNetty.Common;
using Microsoft.Extensions.Logging;
using System;

namespace ConsoleApp1
{
    class Program
    {
        static void Main(string[] args)
        {
            DotNetty.Common.Internal.Logging.InternalLoggerFactory.DefaultFactory.AddConsole();
            ResourceLeakDetector.Level = ResourceLeakDetector.DetectionLevel.Paranoid;

            var logger = DotNetty.Common.Internal.Logging.InternalLoggerFactory.DefaultFactory.CreateLogger(typeof(Program));
            var alloc = ByteBufferUtil.DefaultAllocator;
            for (var i = 0; i < 2; ++i)
            {
                logger.LogInformation("start buf" + i);
                var buf = alloc.Buffer();
                GC.Collect(GC.MaxGeneration, GCCollectionMode.Forced, true);
                for (var j = 0; j < 4; ++j)
                {
                    buf.WriteByte(i);
                    buf.WriteByte(j);
                }
                logger.LogInformation(ByteBufferUtil.HexDump(buf));
                buf.Release();
                logger.LogInformation("end buf" + i);
            }
            logger.LogInformation("The End");

            Console.ReadLine();
        }
    }
}
```
</details>

caozhiyuan:
now when gc , gcnotice will be collect but buffer not write.

add a reference to gcnotice may can slove this problem @yyjdelete  @StormHub .

https://github.com/caozhiyuan/DotNetty/blob/dev/src/DotNetty.Common/ResourceLeakDetector.cs#L81

yyjdelete:
@caozhiyuan I make another patch, tested with `DetectionLevel.Paranoid` and Echo as you described in gitter, can you check if this works for you?
https://github.com/yyjdelete/DotNetty/commit/a67d13e86cd02f5fa74ea9cef19b4f33f5cfbc5f

@StormHub Should the key/value in `gcNotificationMap` be removed when `Rearm` is called? See `CloseFinal`.

StormHub:
[...]

PR #310, opened 2017-11-14
ResourceLeakDetector

StormHub:
1. Leak records target amount and backing off
2. Composite byte buffer copy.
3. Fixed false positive resource leak report sometimes #308 
4. Fixed ByteBufferUtil for string read
5. Removed unused EmptyByteBuffer methods

### DIFF
--- a/src/DotNetty.Buffers/ByteBufferUtil.cs
+++ b/src/DotNetty.Buffers/ByteBufferUtil.cs
@@ -360,7 +360,7 @@ namespace DotNetty.Buffers
                 try
                 {
                     buffer.WriteBytes(src, readerIndex, len);
-                    ArraySegment<byte> ioBuf = src.GetIoBuffer();
+                    ArraySegment<byte> ioBuf = buffer.GetIoBuffer();
                     return encoding.GetString(ioBuf.Array, ioBuf.Offset, ioBuf.Count);
                 }
                 finally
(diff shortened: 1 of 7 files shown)


src/DotNetty.Buffers/ByteBufferUtil.cs
```diff
@@ -360,7 +360,7 @@ public static string DecodeString(IByteBuffer src, int readerIndex, int len, Enc
                 try
                 {
                     buffer.WriteBytes(src, readerIndex, len);
-                    ArraySegment<byte> ioBuf = src.GetIoBuffer();
+                    ArraySegment<byte> ioBuf = buffer.GetIoBuffer();
```
nayato: I assume it means using composite buffers and ToString is broken in 0.4.7, right?

src/DotNetty.Common/ResourceLeakDetector.cs
```diff
@@ -14,17 +14,16 @@ namespace DotNetty.Common
     using DotNetty.Common.Internal;
     using DotNetty.Common.Internal.Logging;
     using DotNetty.Common.Utilities;
-    using Nito;
 
-    public sealed class ResourceLeakDetector
```
nayato: why not seal it?

src/DotNetty.Buffers/ByteBufferUtil.cs
StormHub: Yes, found it in http branch, should be buffer to copy from.

src/DotNetty.Common/ResourceLeakDetector.cs
StormHub: It is not final in netty but nothing extends it for now as far as I know of.

PR #313, opened 2017-11-18
Channel pool

maksimkim:
Motivation:

Provide a generic API and implementation for channel/connection pooling

Modifications:

Add IChannelPool / IChannelPoolMap API and implementations.

Result:

Reusable / Generic pool implementation that users can use.


nayato:
thanks!

PR #338, opened 2018-01-09
Port HA Proxy support from Netty

jsteinich:
This is a port of HA Proxy from that found in the Netty repo.

Changes made:

- I attempted to convert names to match C# conventions, but there may be some more cleanup needed there.
- Some of the referenced internal classes have different names in the two versions.
- Unit tests adapted for different assert methods.
- Functional enums generally converted to constants in regular classes.

I copied comments from the original source, but it would probably be good to reformat them.


master-toma:
Could you please merge conflicts and resubmit PR? I badly need Proxy support :)

jsteinich:
I've updated with changes that I've made in other branches. The ipv6 test is failing, but we've been using it successfully for ipv4. I'll work on fixing ipv6 when I get some more free time.

Issue #328, opened 2017-12-22
Protobuf codecs not sharable


jsteinich:
Is there any reason that ProtobufEncoder and ProtobufDecoder aren't marked as shareable? The java version does have them marked as such.

Issue #196, opened 2017-02-13
Could you please remove the "sealed" for ProtobufEncoder and ProtobufDecoder?


SetoKaiba:
I want to override it to add some feature for my client. My server is in java. And I inherited ProtobufEncoder and ProtobufDecoder to add some feature. But for my client in C#. I can't inherit it.

SetoKaiba:
I can use EmbeddedChannel as a workaround. Anyway, it's great if make codec classes not sealed.

StormHub:
@nayato I think this a good one, your comments? I ran into this issue myself the other day :|

PR #340, opened 2018-01-09
#328 Mark ProtobufDecoder and ProtobufEncoder as shareable

jsteinich:
Make it match Java implementation.

### DIFF
--- a/src/DotNetty.Codecs.Protobuf/ProtobufDecoder.cs
+++ b/src/DotNetty.Codecs.Protobuf/ProtobufDecoder.cs
@@ -22,6 +22,8 @@ namespace DotNetty.Codecs.Protobuf
             this.messageParser = messageParser;
         }
 
+        public override bool IsSharable => true;
+
         protected override void Decode(IChannelHandlerContext context, IByteBuffer message, List<object> output)
         {
             Contract.Requires(context != null);
--- a/src/DotNetty.Codecs.Protobuf/ProtobufDecoder.cs
+++ b/src/DotNetty.Codecs.Protobuf/ProtobufDecoder.cs
@@ -11,7 +11,7 @@ namespace DotNetty.Codecs.Protobuf
     using DotNetty.Transport.Channels;
     using Google.Protobuf;
 
-    public sealed class ProtobufDecoder : MessageToMessageDecoder<IByteBuffer>
+    public class ProtobufDecoder : MessageToMessageDecoder<IByteBuffer>
     {
         readonly MessageParser messageParser;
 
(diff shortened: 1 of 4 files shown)


src/DotNetty.Codecs.Protobuf/ProtobufDecoder.cs
```diff
@@ -22,6 +22,8 @@ public ProtobufDecoder(MessageParser messageParser)
             this.messageParser = messageParser;
         }
 
+        public override bool IsSharable => true;
```
StormHub: The decoder depends on the MessageParser which is typically a specific POCO from proto gen, for instance:
https://github.com/Azure/DotNetty/blob/dev/test/DotNetty.Codecs.Protobuf.Tests/RoundTripTests.cs#L196
In another words, the parser is bound to a message type, this should not be shareable.

src/DotNetty.Codecs.ProtocolBuffers/ProtobufDecoder.cs
```diff
@@ -24,6 +24,8 @@ public ProtobufDecoder(IMessageLite protoType, ExtensionRegistry extensionRegist
             this.extensionRegistry = extensionRegistry;
         }
 
+        public override bool IsSharable => true;
```
StormHub: Same thing as protobuf 3, see
https://github.com/Azure/DotNetty/blob/dev/test/DotNetty.Codecs.ProtocolBuffers.Tests/RoundTripTests.cs#L106
The builder is bound to the message type builder, not really shareable.

src/DotNetty.Codecs.Protobuf/ProtobufDecoder.cs
StormHub: Well, in a specific pipeline, it is shareable since it means that bytes mapped to the protobuf message type message.

src/DotNetty.Codecs.ProtocolBuffers/ProtobufDecoder.cs
StormHub: Same as above, remove the 'sealed'.

src/DotNetty.Codecs.Protobuf/ProtobufDecoder.cs
jsteinich: I also removed 'sealed' from the Encoders per https://github.com/Azure/DotNetty/issues/196

nayato:
I'd prefer to make shareable opt-in, i.e. unseal but not present as Shareable by default; if you know and care to mark it shareable, you can inherit and override then.

jsteinich:
I can switch them to not be shareable by default, but I don't understand why someone wouldn't want them to not be shareable. For starters, the Java version has them as shareable. Perhaps there is something different with how the C# version of protobuf that makes them not function correctly when shared, but at least with protobuf 3, I'm not seeing anything in the source that would cause problems.

PR #352, opened 2018-01-30
Allow encoding and decoding protobuf messages that have no fields

jsteinich:
Since this is a behavior change, we may want to make it configurable, but I'm curious what others think.

I may not have generated the proto cs files with the same parameters that were originally used. If someone has the original commands, I'm happy to regen.

I didn't add composite buffer tests since the resulting buffer only has 1 byte.


src/DotNetty.Codecs.Protobuf/ProtobufDecoder.cs
```diff
@@ -29,10 +29,6 @@ protected override void Decode(IChannelHandlerContext context, IByteBuffer messa
             Contract.Requires(output != null);
 
             int length = message.ReadableBytes;
-            if (length <= 0)
```
StormHub: Personally I don't think this is very good idea because there is no way to tell whether you received something that is empty or nothing is received at all. I ran into similar situation before like this and ended up having 'Identity' message (e.g. all messages with ID, 0 means empty) to ensure the senders intention is unambiguous.

StormHub: In short, received something meaning 'empty' is different from nothing is received at all. On the decoder side, this 'default' behavour might not be desirable.

jsteinich: I certainly agree if someone was trying to just use the raw protobuf data, but I would imagine most use cases have a header or at least length in front of the data. That should indicate whether or not protobuf data is expected.
That said, if we don't want to change the default behavior, perhaps it could be something configurable.

StormHub: Well, I prefer to make them virtual so that you can override them if this is the case. But from comms point of view, I prefer not to have some lower level surprises.

jsteinich: I don't see that much use in overriding the Decode method as it is the functionality.
We could do something like
```c#
int length = message.ReadableBytes;
if (RejectContent(length))
{
    return;
}
...
protected virtual bool RejectContent(int contentLength)
{
    return contentLength <= 0;
}
```
Given that empty messages are just silently dropped with the current behavior, I feel as though it should be documented at a minimum and ideally able to be function.

StormHub: No need for such a simple use case, just override Decode from Decoder which

protected override void Decode(IChannelHandlerContext context, IByteBuffer message, List<object> output) 
{
  if (length <= 0) 
  {
        output.Add("Something empty");
        return;
  }
  base.Decode(context, message, output);
}

Why would we change the base decoder for such an application specific behaviour?

PR #354, opened 2018-02-01
Channel/pipeline writes accept external promise

maksimkim:
Motivation:

Align Write APIs with netty in terms of accepting external promise

Modifications:

* IChannel, IChannelPipeline, IChannelHandlerContext have Write(AndFlush)Async overrrides with promise parameter
* IChannelHandler, IChannelUnsafe WriteAsync converted to synchronous Write which accepts promise
* More extensive void promise support

Result:

* Write APIs are more aligned with netty
* void promise can be used to opt out from promise/task allocation on every write call


src/DotNetty.Transport/Channels/AbstractChannelHandlerContext.cs
```diff
@@ -793,24 +793,46 @@ void InvokeRead()
             }
         }
 
-        public Task WriteAsync(object msg)
+        public Task WriteAsync(object msg) => this.WriteAsync(msg, new TaskCompletionSource());
```
StormHub: Would it be better to have a method on the channel to create new instances of promise (TaskCompletionSource) ?
Once we figured out how to deal with task caching/recyling, we can replace quite easily.
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannel.java#L309
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java#L1055

maksimkim: So the only way to recycle TaskCompletionSource is to replace internal task field on TaskCompletionSource or reset task state itself. Both are not possible (unless we consider reflection). So hiding factory implementation inside channel makes sense if we abstract promise itself and introduce interface which currently will be implemented by TaskCompletionSource only

IPromise
{
 bool TryXXX();
 Task Future;
}

This will include major changes. Do we want to take them now?

StormHub: I don't mean to have any big changes for now.  The main used case for write with task are
await WriteAsync(msg); // Current
WriteAsync(msg).Wait(); // Current
Write(msg, promise) // New 
The current usage just need something awaitable to return (might need change to custom awaitable return type but not breaking change). Therefore, keep the factory method in the channel so that it can be replaced in once place.

maksimkim: So I expect none of the existing use cases will be broken, cause IChannel, IChannelContext, IChannelPipeline all keep existing Task WriteAsync(msg) methods and even new overloads accepting promise return Task as well. It's just IChannelUnsafe and IChannelHandler now have void signatures only, but I assume it's very rare somebody calls Write on these interfaces directly (unless it's dotnetty code).

So the only benefit from promise factory methods on IChannel I see is to be able to change TaskCompletionSource implementation for something better (recyclable/cacheable)), which I'm all for. However, once we come up to that better implementation  not only that factory method signature will have to change, but all other apis accepting TaskCompletionSource as well (cause it will be new type/interface). 

I can do it as part of this PR, but lets decide if we want to replace all direct instantiations of TaskCompletionSource with factory method calls or only those introduced by these changes? Currently it's used quite a lot and not all places have reference to channel object, e.g:

https://github.com/Azure/DotNetty/blob/7aef9c75001ecd926da025540cd2c968510c7dd1/src/DotNetty.Common/Concurrency/RunnableScheduledTask.cs#L11

StormHub: @maksimkim don't get me wrong, the changes look good to me. 
What I had in my mind actually is similar to what netty does, having a factory method on both channel and pipeline, and enable all method to pass in TCS (exactly as you have done), them all TCS (potentially could be some custom promise) can flow through all the way until outbound buffer completes potentially instead of a new TCS almost every step along the way.
What do you think of it?

StormHub: If you assume all tasks are consumed by await, even if the return type changes, it is still not a breaking change, right?

maksimkim: I fully support idea of aligning all async APIs with netty in terms of accepting TCS. But this is separate from aligning with netty in terms of how TCS is created when it has to be created (ctor call vs factory method). We can do these two independently in any order. 

Now my question was if we are doing later (introduce factory method), do we want to switch ALL existing ctor calls to factory method call or only some of them. And if we are doing all, how to deal with cases where channel references is not available.

nayato:
I'm not too worried about breaking changes as long as they are quick to resolve in code. That's the major reason dotnetty is at 0.x version.

src/DotNetty.Transport/Channels/AbstractChannelHandlerContext.cs
nayato: I'm not too worried about breaking changes as long as they are quick to resolve in code. That's the major reason dotnetty is at 0.x version.

StormHub: @maksimkim yeah, let me run some tests on linux and see how it goes. If it does turn out to be worth perusing, I will try to make the promise align closer to netty api.

StormHub: All promises in netty starts with NewPromise on channel including external client code.
e.g.  Promise promise = channel.NewPromise(); channel.Write(data, promise);
This also enables DotNetty to support progress report promise as well.

maksimkim: So. Since DotNetty doesn't have ChannelOutboundInvoker from original netty (btw why?) I introduced newpromise/newpromise(state)/voidpromise on IChannel and IChannelHandlerContext separately.  Also newpromise/newpromise(state) (no voidpromise) was introduced on IEventExecutor. Some components still instantiate TaskCompletionSource directly cause none of 3 factory method owners above are available, e.g. HashedWheelTimer, TlsHandler ctor. Please have a look.

StormHub: @maksimkim This is awesome :) 
Could you please also make the NewPromise directly on IChannelHandlerContext and internally forward to channel NewPromise so that we don't have to do ctx.Channel.NewPromise(), just ctx.NewPromise()

maksimkim: done

StormHub:
Awesome! Now it is possible for everything related to channel/pipeline can either create a new promise or flow through. 👍

maksimkim:
I'm not sure how double virtual method call (IChannelHandlerContext.NewPromise -> IChannel.NewPromise) instead of direct ctor invocation will affect performance, but lets assume those haunting for efficiency will use void promise when possible...

StormHub:
The one on channelContext is not virtual, the one on channel is virtual. If there is nothing overriding them, you can safely seal those.  We can figure what to do next.

maksimkim:
@StormHub, I didn't mean c# language virtuality but rather runtime effect. From CLR perspective it will always be late binding, so no inlining possible.

StormHub:
@maksimkim I see. We will make up for it if it does have some perf implications.

maksimkim:
Decided to include IPromise interface as well while we're here:

1) TaskCompletionSource implementation becomes simpler (no conditional isvoid logic)

2) Now it's possible to port other promise implementations, e.g. https://github.com/netty/netty/blob/73f23c5faa967748bdda826278ac5e63f715ae09/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java#L267

3) Even though EventExecutor in original netty defines generic newPromise<T> implementation it's unlikely we will need IPromise<T> and come up to some alternative to TaskCompletionSource<T> in near future so not introducing it.

StormHub:
@maksimkim Just a note. For the EventExecutor part, netty new promise method by default is just for carrying the executor instance for scheduling.  On DotNetty side, we could use some 'event loop aware' TaskScheduler, on the IPromise impl, we could just look at the scheduler. That was what I thoughts a while back. It is open for discussions, of course.

nayato:
@StormHub we already have that. .NET's approach for this is `ThreadScheduler.Current` which is captured  for continuations unless you explicitly opt out explicitly (e.g. with `.ConfigureAwait(false)`).

StormHub:
@nayato you mean the one on event loop thread start task?

StormHub:
It could be assisted with the promise

maksimkim:
Closing in favor of https://github.com/dotnet/corefx/issues/27445

PR #355, opened 2018-02-02
According to MSDN, "Using Shutdown on a connectionless Socket is not …

SylphiaWindy:
…recommended"

and the fact is calling Shutdown() on  an udp socket will cause an exception (socket not connected)
So we should simply call Close() then Dispose().
But there's no Close() method in dotnet core 1.1, so just an Dispose() would be enough.

P.S. I can't get the example QuoteOfTheMoment working with netCore 1.1 (1.1.7) on macOS 10.13, "Exception: System.PlatformNotSupportedException: This platform does not support packet information for dual-mode sockets", disable dual mode of the socket gives a invalid argument socket error (22).

### DIFF
--- a/src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
+++ b/src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
@@ -97,7 +97,9 @@ namespace DotNetty.Transport.Channels.Sockets
         {
             if (this.TryResetState(StateFlags.Open | StateFlags.Active))
             {
-                this.Socket.Shutdown(SocketShutdown.Both);
+#if !NETSTANDARD1_3                
+                this.Socket.Close();
+#endif                
                 this.Socket.Dispose();
             }
         }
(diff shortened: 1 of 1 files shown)


SylphiaWindy:
Maybe just use ReceiveAsync() instead of ReceiveFromAsync() ?

src/DotNetty.Transport/Channels/Sockets/SocketDatagramChannel.cs
```diff
@@ -97,7 +97,9 @@ protected override void DoClose()
         {
             if (this.TryResetState(StateFlags.Open | StateFlags.Active))
             {
-                this.Socket.Shutdown(SocketShutdown.Both);
+#if !NETSTANDARD1_3
```
nayato: Dispose calls Close internally (if it wasn't called before) so no need to add explicit Close here.

SylphiaWindy: Agreed.

PR #361, opened 2018-02-21
Change default socket creation to handle OS where IPv6 not supported.

darobs:
Related to .Net Core issue: https://github.com/dotnet/corefx/issues/26664

.Net Sockets default to IPv6 with DualMode=true on creation, but on Linux systems where IPv6 is not present (disabled in the kernel), the attempt to create the socket in this mode throws a libuv error of EAFNOSUPPORT.

To work around this, I changed the default Socket creation to set the address family to IPv6 and DualMode=true only if `Socket.OSSupportsIPv6` is true.  There should be no net effect to the behavior on the systems which support IPv6.


StormHub:
Should we do this similar to Netty, introducing options https://github.com/netty/netty/blob/81f9de423c03ec4538f1ba04b89ae049959b698c/resolver/src/main/java/io/netty/resolver/ResolvedAddressTypes.java
to handle this kind of issues?

I think this might have to involve the name resolver:
https://github.com/Azure/DotNetty/blob/dev/src/DotNetty.Transport/Bootstrapping/DefaultNameResolver.cs

/cc @nayato

PR #369, opened 2018-03-07
Buffer read/write/get/set strings.

StormHub:
Motivation:
Implement commonly used string-based operations (originally Netty CharSequence) for byte buffers.

Modifications:
Added read/write/get/set strings methods to byte buffers.
Also reduced unsafe pin/unpins for byte operations.

Result:
String-based operations are supported in DotNetty

### DIFF
--- a/src/DotNetty.Buffers/ByteBufferUtil.cs
+++ b/src/DotNetty.Buffers/ByteBufferUtil.cs
@@ -12,6 +12,9 @@ namespace DotNetty.Buffers
 
     public static class ByteBufferUtil
     {
+        const char WriteUtfUnknown = '?';
+        static readonly int MaxBytesPerCharUtf8 = Encoding.UTF8.GetMaxByteCount(1);
+
         static readonly IInternalLogger Logger = InternalLoggerFactory.GetInstance(typeof(ByteBufferUtil));
 
         public static readonly IByteBufferAllocator DefaultAllocator;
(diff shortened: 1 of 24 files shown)


src/DotNetty.Buffers/ByteBufferUtil.cs
```diff
@@ -295,6 +297,149 @@ static int LastIndexOf(IByteBuffer buffer, int fromIndex, int toIndex, byte valu
...
+                    {
+                        buffer._SetByte(writerIndex++, WriteUtfUnknown);
+                        continue;
+                    }
+                    char c2;
+                    try
+                    {
+                        // Surrogate Pair consumes 2 characters. Optimistically try to get the next character to avoid
+                        // duplicate bounds checking with charAt. If an IndexOutOfBoundsException is thrown we will
+                        // re-throw a more informative exception describing the problem.
```
nayato: where does this re-throwing happens?

src/DotNetty.Buffers/ByteBufferUtil.cs
```diff
@@ -295,6 +297,149 @@ static int LastIndexOf(IByteBuffer buffer, int fromIndex, int toIndex, byte valu
...
+                    buffer._SetByte(writerIndex++, (byte)(0xe0 | (c >> 12)));
+                    buffer._SetByte(writerIndex++, (byte)(0x80 | ((c >> 6) & 0x3f)));
+                    buffer._SetByte(writerIndex++, (byte)(0x80 | (c & 0x3f)));
+                }
+            }
+
+            return writerIndex - oldWriterIndex;
+        }
+
+        public static int Utf8MaxBytes(string value) => Encoding.UTF8.GetMaxByteCount(value.Length);
```
nayato: I'm ok with it if it's internal. Otherwise, let's not publish APIs unrelated to a subject (Buffer mgmt here)

StormHub: Well, if that happens L316 will through buffer index out of range

src/DotNetty.Buffers/ByteBufferUtil.cs
StormHub: Well, if that happens L316 will through buffer index out of range

src/DotNetty.Buffers/ByteBufferUtil.cs
StormHub: I changed it to internal, in netty it is public but that is for CharSequence, not string.

alexmg:
An improvement to the API and performance. Nice work @StormHub! 👍

PR #370, opened 2018-03-07
close channel on write failure

nayato:


nayato:
@maksimkim PTAL

nayato:
@StormHub your opinion? we've been missing this from the beginning as 5.0 alpha of netty had it removed (per deprecation policy). Now I don't really see why one might want to keep their channel open in half-broken state but I might be missing smth here.

StormHub:
@nayato It should be ok. The only difference is that Netty catches all exception and close immediately if it is java IOException (IO failures), otherwise Netty will attempt to shutdown the channel (e.g. shutdown  socket or close native transport gracefully), only if that fails, forcibly close the channel.
The channel will be closed either way, the  different is that should we give the channel a chance to gracefully close if the exception is not a IO failure.

nayato:
@StormHub the issue is, there's no single exception in .net to catch. there is SocketException that is specific to sockets, there's IO exception that is unrelated to networking errors in .NET, libuv also has own exception types, I believe. On the other hand, I don't see what else could cause a failure writing into a channel directly (that's unsafe-level API).

StormHub:
@nayato yep, java IOException means network failure, in .net it actually means file related stuff. Having said that, you can catch SocketException for Socket based transport :)

nayato:
well, then we'd need to develop a universal contract for transports to comm intent, akin IOException in java. I think SocketException is too specialized to use it in that capacity.

StormHub:
yep, either way I don't think it is a big deal for now.

PR #375, opened 2018-03-12
Allocation-free Write(AndFlush)Async using ValueTask

maksimkim:
Motivation:

corefx 2.1 will introduce ability to implement allocation free async awaitables using existing ValueTask struct and new IValueTaskSource interface (https://github.com/dotnet/corefx/issues/27445). PR adopts this new feature for write apis to avoid unnecessary Task allocation on every WriteAsync call.

ValueTask backed by IValueTaskSource by design has lots of limitations (see corefx issue for details). 

In this PR IValueTaskSource is mostly implemented by recyclable objects (ChannelOutboundBuffer.Entry, PendingWriteQueue.PendingWrite, AbstractChannelHandlerContext.WriteTask, etc). It means than once completed it is immediately recycled which has certain implication on how it can be used inside and outside DotNetty code:

* Once ValueTask is returned by WriteAsync call it should be either i**mmediately awaited** or **immediately converted to Task** or **never touched again** (e.g. storing it as a class field is not allowed).
Not following this restriction can't be verified by compiler and can cause runtime error (if it was recycled after completion) or unpredictable result (if it was reactivated from pool for different async operation).

* WriteAndFlushAsync method doesn't follow guideline above (it calls WriteAsync then Flush and then returns ValueTask to a caller) that is why new overload of WriteAndFlushAsync is introduced. 
New notifyComplete bool parameter defines if caller is interested in completion notification (for the penalty of Task allocation) or not. In later case completed ValueTask is returned from WriteAndFlushAsync.

Modifications:

* Corresponding projects reference new System.Threading.Tasks.Extensions package version from myget dotnet feed. Assumption is that by the time PR is reviewed new version is available on nuget.org.
* New promise hierarchy is introduced (IPromise, AbstractPromise, AbstractRecyclablePromise) to provide similar capabilities as TaskCompletionSource but for ValueTask/IValueTaskSource pattern.
* WriteAsync/WriteAndFlushAsync return ValueTask
* WriteAndFlushAsync has new overload with notifyComplete bool parameter to opt out of Task allocation

Result:

* If calling code is not interested in awaiting write completion no TaskCompletionSource/Task allocation happen (DotNetty.Transport.Tests.Performance.Sockets.TcpChannelPerfSpecs+TcpChannel_Duplex_Throughput_10_messages_per_flush has twice less Gen0 collections).


examples/Telnet.Server/TelnetServerHandler.cs
```diff
@@ -35,10 +36,10 @@ protected override void ChannelRead0(IChannelHandlerContext contex, string msg)
                 response = "Did you say '" + msg + "'?\r\n";
             }
 
-            Task wait_close = contex.WriteAndFlushAsync(response);
+            ValueTask wait_close = contex.WriteAndFlushAsync(response);
```
nayato: waitClose?

src/DotNetty.Codecs/MessageToByteEncoder.cs
```diff
@@ -52,13 +53,13 @@ public override Task WriteAsync(IChannelHandlerContext context, object message)
...
-            catch (EncoderException e)
+            catch (EncoderException)
             {
-                return TaskEx.FromException(e);
+                throw;
             }
             catch (Exception ex)
             {
-                return TaskEx.FromException(new EncoderException(ex));
+                throw new EncoderException(ex);
```
nayato: throwing synchronously in async method is a bad practice. Can we return failure in ValueTask?

nayato (changes requested):

src/DotNetty.Codecs/MessageToByteEncoder.cs
maksimkim: Any reason it was diverged from original netty in a first place:
https://github.com/netty/netty/blob/0692bf1b6a6764dee22132630bef7ee2f850771e/codec/src/main/java/io/netty/handler/codec/MessageToByteEncoder.java#L125

nayato: and what happens to that exception?

maksimkim: Presumable it baubles up all the way to a caller

maksimkim: However ChannelHandlerContext always converts it to failed future:
https://github.com/netty/netty/blob/74f24a5c19f8f351e9c6a7a84bdd9fbcc7a07ada/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java#L740

So probably context is assumed to be higher level exception handler here.

nayato: so, pro: we're collecting stack trace which might be useful debugging real issue
cons: we're collecting stack trace which is costly (prohibitively so in some scenarios).
not sure what is worse here. I'd guess if we wanted to return a failed future here it would be a valuetask wrapping failed Task, right?

maksimkim: main pro I see: we're close to original source code so it won't cause unexpected behavior when porting code based on this encoder/decoder infra in future.

For failed future yes, Exception -> Task -> ValueTask. Stephen Toub explicitly said that exceptions are already expensive enough and making ValueTask reference exception instance directly doesn't make sense.

src/DotNetty.Common/Utilities/TaskEx.cs
```diff
@@ -72,14 +77,84 @@ public static void LinkOutcome(this Task task, TaskCompletionSource taskCompleti
...
+                    task.ContinueWith(LinkOutcomePromise, promise, TaskContinuationOptions.ExecuteSynchronously);
                     break;
             }
         }
 
+        public static async void LinkOutcome(this ValueTask future, IPromise promise)
+        {
+            try
+            {
+                await future;
```
maksimkim: AbstractPromise now can flow both execution and synchronization context. What is our context management story? 
* Can we use await in dotnetty
* If we can, should it always be ConfigureAwait false?
* Should we suppress execution context?

src/DotNetty.Codecs/MessageToByteEncoder.cs
StormHub: @nayato @maksimkim If we assume Exceptions are rare, it would ok to accept the perf hit.

src/DotNetty.Common/Utilities/TaskEx.cs
nayato: if you do `.ConfigureAwait(false)`, you would erase notion of custom task scheduler that dispatches continuation to this same event loop.

maksimkim: So in case of ValueTask it's IValueTaskSource implementation responsibility to execute continuation. That implementation presumably not aware of any task scheduler.

src/DotNetty.Handlers/Streams/ChunkedWriteHandler.cs
```diff
@@ -228,54 +229,62 @@ void DoFlush(IChannelHandlerContext context)
...
-                            TaskContinuationOptions.ExecuteSynchronously);
+                                await future;
+                                promise.Progress(input.Progress, input.Length);
+                            }
+                            catch(Exception ex)
+                            {
+                                CloseInput((IChunkedInput<T>)promise.Message);
+                                promise.TrySetException(ex);
+                            }
+                        }
```
maksimkim: @StormHub , I've aligned behavior with netty here: progress() never completes promise
https://github.com/netty/netty/blob/8d78893a76a8c14c18b304ffe1fce2dc43ed4206/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java#L278

Any reason it was ported in different way?

StormHub: @maksimkim I don't think there is a particular reason. Let me double check.

build.cake
```diff
@@ -50,7 +50,7 @@ Task("Restore-NuGet-Packages")
   .Description("Restores dependencies")
   .Does(() =>
 {
-  DotNetCoreRestore();
+  DotNetCoreRestore(new DotNetCoreRestoreSettings { ConfigFile = ".nuget\\nuget.config" });
```
nayato: do we still need this? and change to nuget.config above?

src/DotNetty.Transport/Channels/IChannelHandlerContext.cs
```diff
@@ -67,11 +67,13 @@ public interface IChannelHandlerContext : IAttributeMap
 
         IChannelHandlerContext Read();
 
-        Task WriteAsync(object message); // todo: optimize: add flag saying if handler is interested in task, do not produce task if it isn't needed
+        ValueTask WriteAsync(object message); // todo: optimize: add flag saying if handler is interested in task, do not produce task if it isn't needed
 
         IChannelHandlerContext Flush();
 
         Task WriteAndFlushAsync(object message);
```
maksimkim: Since WriteAndFlushAsync is consequent WriteAsync and Flush, and later can cause former to complete synchronously, ValueTask from WriteAsync backend by recyclable future is always materialized in full blown Task. So I kept Task as return type. Assumption is that those who want to do write+flush in non-allocating fire-and-forget mode  call overload and pass _false_ in notifyComplete. In that case overload always returns completed ValueTask.

src/DotNetty.Handlers/Streams/ChunkedWriteHandler.cs
StormHub: @maksimkim I think I did it wrong. Netty progress never completes the promise.

StormHub: @maksimkim also note that success also calls progress
https://github.com/netty/netty/blob/4f982be91e6d61c7c7a9d8d63115e0ebe65b2774/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java#L350

src/DotNetty.Codecs.Http/HttpServerUpgradeHandler.cs
```diff
@@ -254,34 +254,28 @@ bool Upgrade(IChannelHandlerContext ctx, IFullHttpRequest request)
...
-                        // Release the event if the upgrade event wasn't fired.
-                        upgradeEvent.Release();
-                    }
-                }, TaskContinuationOptions.ExecuteSynchronously);
+            try
+            {
+                Task writeTask = ctx.WriteAndFlushAsync(upgradeResponse);
+                
+                // Perform the upgrade to the new protocol.
+                this.sourceCodec.UpgradeFrom(ctx);
```
maksimkim: @StormHub, check this out. I change it to align with netty. So upgrade doesn't wait write to complete. 
https://github.com/netty/netty/blob/00afb19d7a37de21b35ce4f6cb3fa7f74809f2ab/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java#L324

src/DotNetty.Codecs.Http/WebSockets/Extensions/WebSocketServerExtensionHandler.cs
```diff
@@ -88,31 +89,33 @@ public override Task WriteAsync(IChannelHandlerContext ctx, object msg)
...
-                };
-
                 if (headerValue != null)
                 {
-                    response.Headers.Set(HttpHeaderNames.SecWebsocketExtensions, headerValue);
+                    responseHeaders.Set(HttpHeaderNames.SecWebsocketExtensions, headerValue);
                 }
+
+                Task task = base.WriteAsync(ctx, msg).AsTask();
+                task.ContinueWith(this.upgradeCompletedContinuation, ctx, TaskContinuationOptions.ExecuteSynchronously);
```
maksimkim: @StormHub, check this out. Aligned with netty
https://github.com/netty/netty/blob/00afb19d7a37de21b35ce4f6cb3fa7f74809f2ab/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/extensions/WebSocketServerExtensionHandler.java#L118

PR #379, opened 2018-03-14
Better inline some methods in AbstractByteBuffer

yyjdelete:
This may makes call to GetByte(int) 25%~50% faster


StormHub:
@yyjdelete I have some inline changes in #369

yyjdelete:
Will see again after it's merged.

StormHub:
@yyjdelete I will a recheck after it gets merged. BTW: Benchmark definitely show improvements both in GetByte and SetByte for sure!

PR #380, opened 2018-03-18
Better inline AbstractByteBuffer.CheckIndex0

yyjdelete:
That make `Get/Set` about 50% faster in netcoreapp/netfx(RyuJIT only, the old JIT is not tested), and almost the same as it used to be in mono.

Not sure the extra reason, maybe it's too long to be inlined.

### DIFF
--- a/src/DotNetty.Buffers/AbstractByteBuffer.cs
+++ b/src/DotNetty.Buffers/AbstractByteBuffer.cs
@@ -1330,7 +1330,7 @@ namespace DotNetty.Buffers
         {
             if (MathUtil.IsOutOfBounds(index, fieldLength, this.Capacity))
             {
-                ThrowHelper.ThrowIndexOutOfRangeException($"index: {index}, length: {fieldLength} (expected: range(0, {this.Capacity}))");
+                ThrowHelper.ThrowIndexOutOfRangeException("index: {0}, length: {1} (expected: range(0, {2}))", index, fieldLength, this.Capacity);
             }
         }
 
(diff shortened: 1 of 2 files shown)


yyjdelete:
It's interesting that seems it's not related to inline, it's even slower with `MethodImplOptions.NoInlining`. 
And slower than `throw new IndexOutOfRangeException($"srcIndex: {srcIndex}, length: {length} (expected: range(0, {srcCapacity}))")` or `throw ThrowHelper.CreateIndexOutOfRangeException($"srcIndex: {srcIndex}, length: {length} (expected: range(0, {this.Capacity}))")`


Not sure what's happened, and maybe something related to the JIT compiler.

yyjdelete:
@StormHub @nayato 
Maybe we should move all `string.Format` into `ThrowHelper`( and optionally splitted to an `ThrowXxx()` and an `GetXxx()`) or move  `throw` out of `ThrowHelper` in all those cases?  

See https://github.com/dotnet/coreclr/issues/17024

StormHub:
We should do something similar to https://github.com/gfoidl/Common/commit/87b9a54ff63d07e6812f7256eab07527304a4ee8

StormHub:
Specifically like this one
https://github.com/gfoidl/Common/blob/87b9a54ff63d07e6812f7256eab07527304a4ee8/src/Microsoft.Extensions.Primitives/StringSegment.cs#L594

StormHub:
@yyjdelete @nayato let me change the inlining style and try benchmark again

StormHub:
BenchmarkDotNet InliningDiagnoser reported this:

Inliner: DotNetty.Microbench.Buffers.PooledByteBufferBenchmark.CheckIndex - instance void  ()
Inlinee: DotNetty.Buffers.AbstractByteBuffer.CheckIndex0 - instance void  (int32,int32)
Fail Reason: has ldstr VM restriction

nayato:
@StormHub what's going on in here? :)

StormHub:
@nayato I have added the InliningDiagnosor and working out a way to ensure it is inlined most of the time.

PR #382, opened 2018-03-21
Netstandard2 multi target

jsteinich:
Adds netstandard2.0 as an additional target built.
Updates Microsoft.Extensions.Logging to 2.0.0 when using the netstandard2.0 target.

The motivation behind this is to fix a dependency conflict with MEL when using a different library with DotNetty that requires 2.0.0 and MEL 2.0.0 requires netstandard2.0.

### DIFF
--- a/src/DotNetty.Handlers/Tls/TlsHandler.cs
+++ b/src/DotNetty.Handlers/Tls/TlsHandler.cs
@@ -665,8 +665,8 @@ namespace DotNetty.Handlers.Tls
             int inputLength;
             TaskCompletionSource<int> readCompletionSource;
             ArraySegment<byte> sslOwnedBuffer;
-#if NETSTANDARD1_3
-            int readByteCount;
+#if NETSTANDARD1_3 || NETSTANDARD2_0
+			int readByteCount;
 #else
             SynchronousAsyncResult<int> syncReadResult;
             AsyncCallback readCallback;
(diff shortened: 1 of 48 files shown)


StormHub:
Add netstandard 2.0 simply for logging does not add much value in my opinion. I would rather prefer a full migration to netstandard 2.0 and depreciate netstandard1.3. 
/cc @nayato @maksimkim any opinion on this?

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -710,8 +710,8 @@ public void ExpandSource(int count)
 
                 ArraySegment<byte> sslBuffer = this.sslOwnedBuffer;
 
-#if NETSTANDARD1_3
-                this.readByteCount = this.ReadFromInput(sslBuffer.Array, sslBuffer.Offset, sslBuffer.Count);
+#if NETSTANDARD1_3 || NETSTANDARD2_0
+				this.readByteCount = this.ReadFromInput(sslBuffer.Array, sslBuffer.Offset, sslBuffer.Count);
                 // hack: this tricks SslStream's continuation to run synchronously instead of dispatching to TP. Remove once Begin/EndRead are available.
```
yyjdelete: Begin/End Read/Write in already added in NETSTANDARD2_0, so maybe override them in this class(and the one in test)?

StormHub: @yyjdelete Exactly! There are things need to be cleaned up specifically for netstandard 2.0 before we officially support it. Supporting both 1.3 and 2.0 would make some part of code much harder to maintain. Plus, there are so many things we can take advantage of in netstandard 2.0

tedvanderveen:
Great PR! Thanks. Any ETA?

src/DotNetty.Handlers/Tls/TlsHandler.cs
jsteinich: I can attempt to help with that transition, but I'm not all that familiar with the internals of DotNetty or what specifically netstandard 2.0 added (other than a lot).
To me it seems like a larger project that should be broken out into pieces (perhaps starting with multi-target as the community works on the rest). It also seems that DotNetty is due for a release before that project gets started (would be really nice to get that http pr merged).

PR #392, opened 2018-04-29
Fix DiskFileUpload

hey-red:
Fixed disk attributes 519f2d7
We should be reset stream position, otherwise DiskAttribute.Value property produces empty array as string.
Similar problem for disk files 34c9a1e
818208b After using RenameTo we don't want that file will be removed when decoder starts cleanup.


hey-red:
Also, maybe replace [this](https://github.com/Azure/DotNetty/blob/dev/src/DotNetty.Codecs.Http/Multipart/AbstractDiskHttpData.cs#L272) to CopyTo?

StormHub:
lol, http is not release yet. Would you please show how to reproduce this?

hey-red:
@StormHub I'm just trying to implement simple upload handler https://gist.github.com/hey-red/35345014eec62c224450fa77f8f498f4
Server same as https://github.com/Azure/DotNetty/tree/dev/examples/HttpServer

nayato:
@StormHub can you comment on this? I don't really have enough context to review this properly

PR #398, opened 2018-06-14
Fix that an canceled ScheduledTask can still be executed in rare case

yyjdelete:


yyjdelete:
The CI script is broken?

----

Still busy now, maybe will pickup patch from #374 about a week latter.

PR #413, opened 2018-08-07
Fix socket connecting synchronously

varunpuranik:
This PR fixes 2 issues - 
1. If Socket.ConnectAsync returns synchronously, then we cannot assume that the connection succeeded. The connection needs to be validated. 
2. When Fulfilling the connect promise FireChannelActive should be called irrespective of whether the Promise is set. This is because the Promise is set only if the Socket.ConnectAsync returns asynchronously.


CIPop:
.NET runtime versions containing the following feature are affected: https://github.com/dotnet/corefx/issues/4900

Issue #415, opened 2018-08-18
Converting IByteBuffer to byte[] throws IndexOutOfRangeException with UnpooledByteBufferAllocator


varunpuranik:
We use the following code to convert an IByteBuffer to a byte[] - 
```
        public byte[] ToByteArray(IByteBuffer byteBuffer)
        {
            if (!byteBuffer.IsReadable())
            {
                return new byte[0];
            }
            using (var stream = new ReadOnlyByteBufferStream(byteBuffer, false))
            {
                var memoryStream = new MemoryStream();
                stream.CopyTo(memoryStream);
                byte[] bytes = memoryStream.ToArray();
                return bytes;
            }
        }
```

But this throws an IndexOutOfRangeException with the following callstack - 

at DotNetty.Buffers.UnpooledUnsafeDirectByteBuffer.Addr(Int32 index) in E:\workspace\DotNetty\src\DotNetty.Buffers\UnpooledUnsafeDirectByteBuffer.cs:line 339
   at DotNetty.Buffers.UnpooledUnsafeDirectByteBuffer.GetBytes(Int32 index, Byte[] dst, Int32 dstIndex, Int32 length) in E:\workspace\DotNetty\src\DotNetty.Buffers\UnpooledUnsafeDirectByteBuffer.cs:line 211
   at DotNetty.Buffers.AbstractByteBuffer.ReadBytes(Byte[] destination, Int32 dstIndex, Int32 length) in E:\workspace\DotNetty\src\DotNetty.Buffers\AbstractByteBuffer.cs:line 998
   at DotNetty.Buffers.ReadOnlyByteBufferStream.Read(Byte[] output, Int32 offset, Int32 count) in E:\workspace\DotNetty\src\DotNetty.Buffers\ReadOnlyByteBufferStream.cs:line 46
   at System.IO.Stream.CopyTo(Stream destination, Int32 bufferSize)
   at DotnettyTest1.ByteBufferConverter.ToByteArray(IByteBuffer byteBuffer) in C:\Users\vpuranik.REDMOND\Documents\Visual Studio 2017\Projects\DotnettyTest1\DotnettyTest1\ByteBufferConverter.cs:line 34
   at DotnettyTest1.Program.ByteBufferTest1() in C:\Users\vpuranik.REDMOND\Documents\Visual Studio 2017\Projects\DotnettyTest1\DotnettyTest1\Program.cs:line 26
[...]

PR #438, opened 2018-10-21
Fix #436, #415

### LINKED ISSUE #436: Do some fix for UnpooledUnsafeDirectByteBuffer
1. Fix the possible data mismatch and `AccessViolationException` for `UnpooledUnsafeDirectByteBuffer` in netcoreapp1.x
    >Apply workaround of dotnet/coreclr#17071 for netcoreapp1.x
    >See the issue at Azure/azure-iot-sdk-csharp#604. DotNetty `0.6.0` already prefer safe version of buffers, but it's also ok to fix it.
    >https://gist.github.com/yyjdelete/210da6dfb09a0d588b567e5deada6847 , and try with `dotnet run -c Release -f netcoreapp1.1`

2. Fix the `IndexOutOfRangeException` of `UnpooledUnsafeDirectByteBuffer`
    >#415, And when try to read 0 bytes at the end of the buffer(or capacity of buffer is zero).
   >Possible also for , and `GetPinnableMemoryAddress()

Will do this when I'm free.

yyjdelete:


yyjdelete:
The new `AsRef` return an ref version of `IntPtr.Zero` when `offset >= array.Length`(or array.Length is 0) and can be fixed to `(void*)0`, which similar to [Span.GetPinnableReference()](https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Span.Fast.cs#L161) does. It's safe since if `AbstractByteBuffer.CheckIndex` works, the expected Length is always zero in this case, so nothing will be done actually. 
But the asserts may not pass if the pointer are used by some native code outside(DllImport or etc), and some extra works must be done when call Add.

Maybe System.Runtime.InteropServices.MemoryMarshal.GetReference(new ReadOnlySpan<byte>(source, offset, 0)) works better than the new version of AsRef in such case, as it can always return an
actually unsafe pointer of the array, even if it's length is 0. But it introduce 3 new dependencies and maybe slower on some old platforms which dosen't has native support of Span<T>.
See the difference from `https://github.com/dotnet/corefx/issues/32099`, `https://github.com/dotnet/corefx/issues/28969`

System.Memory
|-System.Buffers
|-System.Numerics.Vectors
|-System.Runtime.CompilerServices.Unsafe

PR #459, opened 2019-01-04
Fix ParseMachineId

### LINKED ISSUE #458: When specifying the netty machine id, parsemachineid does not work
I dont believe it is possible for this code to work:

https://github.com/Azure/DotNetty/blob/580bde0415a50e570eb9a40b08142d28569ec798/src/DotNetty.Transport/Channels/DefaultChannelId.cs#L119

onionhammer:
Fix issue parsing manual machine id; fixes #458 - second parameter of `Substring` is 'length' not 'index'


onionhammer:
@nayato When will this be released?

Issue #492, opened 2019-05-22
chunkedWriteHandler does not work when use chunkedStream(fileStream)


cocosip:
I use chunkedWriteHandler to send file.
```c#
 var stream = new ChunkedStream(request.Stream, 1024 * 1024 * 10);
 _channel.WriteAndFlushAsync(stream).Wait();
```
when the request.Stream is `MemoryStream` , it works well. 
but when the request.Stream is `FileStream` , it seem not work.

cocosip:
I know the cause of the problem.
In `chunkedStream`  ,the method `public IByteBuffer ReadChunk(IByteBufferAllocator allocator)`  have a code:
```c#
int count = buffer.SetBytesAsync(buffer.WriterIndex, this.input, readChunkSize, CancellationToken.None).Result;
```
in buffer implements such as  `PooledHeapByteBuffer`, `CompositeByteBuffer`, `UnpooledHeapByteBuffer`, `UnpooledHeapByteBuffer`, the method `SetBytesAsync` is `async` and it not use `ConfigureAwait(false)` ,so it cause **deadlock**

caozhiyuan:
@StormHub pelease check chunkedStream  buffer.SetBytesAsync .Result

PR #493, opened 2019-05-23
fix bug #492,async cause deadlock.

cocosip:
Signed-off-by: virtual <1185513330@qq.com>


nayato:
.ConfigureAwait(false) here will cause fields of the buffer to be updated on a different thread which presents data race issues. As the issue has already been closed I don't see a reason to keep this PR around.

PR #497, opened 2019-06-27
Port all properties io.netty.recycler.* and related logic

yyjdelete:
That help `ThreadLocalPool` not cache too much objects when huge sudden transfer, or when alloc and release happen on different thread.

The PR is already finished and no test failed. But I'm not sure the mean of comments below, and not sure whether it also works or not for dotnet.
https://github.com/Azure/DotNetty/pull/497/commits/3840311361fb42974f4f63387b265011014ee018#diff-629c893e2d4b4fee364a76632af183b6R230

### DIFF
--- a/src/DotNetty.Common/ThreadLocalPool.cs
+++ b/src/DotNetty.Common/ThreadLocalPool.cs
@@ -3,7 +3,10 @@
 
 namespace DotNetty.Common
 {
+    using DotNetty.Common.Internal;
+    using DotNetty.Common.Internal.Logging;
     using System;
+    using System.Diagnostics;
     using System.Diagnostics.Contracts;
     using System.Runtime.CompilerServices;
     using System.Threading;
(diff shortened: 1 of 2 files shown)


vovgou:
There is a bug in this PR, which will cause the pooled object released in another thread to not be recycled correctly.
Please use the following code to test it. The object pool will become invalid after reusing the first few objects.


```
using DotNetty.Common;
using System;
using System.Collections.Generic;
using System.Threading;
using System.Threading.Tasks;

public class RecyclerExample
{
    List<PooledObject> pool = new List<PooledObject>();
    object _lock = new object();
    bool running = true;
    void DoTest()
    {
        Environment.SetEnvironmentVariable("io.netty.recycler.ratio", "0");

        Task.Run(() =>
        {
            while (running)
            {
                lock (_lock)
                {
                    if (pool.Count > 5)
                    {
                        PooledObject obj = pool[0];
                        pool.RemoveAt(0);
                        obj.Release();
                    }
                    else
                    {
                        Monitor.Wait(_lock);
                    }
                }
            }
        });

        Task.Run(() =>
        {
            while (running)
            {
                try
                {
                    Thread.Sleep(100);
                }
                catch (Exception) { }

                var obj = PooledObject.NewInstance();
                Console.WriteLine("id:{0}", obj.ID);
                pool.Add(obj);
                lock (_lock)
                {
                    Monitor.PulseAll(_lock);
                }
            }
        });
    }
}

public class PooledObject
{
    static readonly ThreadLocalPool<PooledObject> Recycler = new ThreadLocalPool<PooledObject>(handle => new PooledObject(handle), 256);
    static int sequence = 0;
    public static PooledObject NewInstance()
    {
        PooledObject message = Recycler.Take();
        return message;
    }

    private ThreadLocalPool.Handle recyclerHandle;
    private int id;
    private PooledObject(ThreadLocalPool.Handle handle)
    {
        this.recyclerHandle = handle;
        this.id = sequence++;
    }

    public int ID => this.id;

    public void Release()
    {
        if (this.recyclerHandle != null)
        {
            this.recyclerHandle.Release(this);
        }
    }

    public override string ToString()
    {
        return string.Format("PooledObject:{0}", ID);
    }
}

```

Issue #505, opened 2019-07-31
mark GetUnadjustedFrameLength (LengthFieldBasedFrameDecoder.cs) as virtual


dimdeli:
We faced an implementation with a **6 bytes ASCII encoded** length field header.

So it would be pretty cool to inherit from LengthFieldBasedFrameDecoder and **have the ability just to override this method** in order to decode the length field appropriately!

PR #506, opened 2019-07-31
#505 - mark GetUnadjustedFrameLength as virtual

dimdeli:


dimdeli:
In Java, all non-static methods are by default "virtual functions." Only methods marked with the keyword final, which cannot be overridden, along with private methods, which are not inherited, are non-virtual.

https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java#L452

sunliusi:
can you publish a new version?

PR #507, opened 2019-08-01
Fix #498

### LINKED ISSUE #498: Id conflict for AttributeKey<T> with different types
Unlike java, which does generic type erase; `AttributeKey<Class1>` and `AttributeKey<Class2>` are 2 different types in dotnet, that mean they will use 2 separate `AttributeConstantPool`, and can both the same `Id`.

That will make `DefaultAttributeMap` inefficient(still work, just inefficient)
, because it always use bucket-1, and never use 0,2,3 when you use more than 1 `AttributeKey<T>` and always with different types.

Maybe `AttributeKey<T>` needed to be separated to 2 types(like System.Tuple and System.Tuple<T1,...>), use an new utils class `static class AttributeKey` to hold all static methods and fields(or just fields for binary compatibility), and use the old `sealed class AttributeKey<T> : AbstractConstant<AttributeKey<T>>` for the data class itself.

~~***NOTE: *** It's always not safe for binary compatibility, since `AttributeKey<T>.Pool` is an public static field, and may be used by libs.~~
Can be marked as Obsolete and set `AttributeKey<T>.Pool = AttributeKey.Pool`

https://github.com/Azure/DotNetty/blob/72deeb9a09627763ef82e543109a8a85a5334095/src/DotNetty.Common/Utilities/AttributeKey.cs#L15-L17
https://github.com/Azure/DotNetty/blob/72deeb9a09627763ef82e543109a8a85a5334095/src/DotNetty.Common/Utilities/DefaultAttributeMap.cs#L35

yyjdelete:
More details can be see in #498 

In fact, `AttributeKey` can also be public and move all static method into the new static class(In fact `AttributeKey<T>.Exists(string)` never do check for `typeof(T)`), like `Tuple.Create<T1, T2>()` and `Tuple<T1, T2>` does. But it will be an breaking change.

### DIFF
--- a/src/DotNetty.Common/Utilities/AttributeKey.cs
+++ b/src/DotNetty.Common/Utilities/AttributeKey.cs
@@ -5,6 +5,18 @@ namespace DotNetty.Common.Utilities
 {
     using System;
 
+    internal static class AttributeKey
+    {
+        // Keep the instance of AttributeConstantPool out of generic classes, to make it an really singleton for different generic types.
+        // see https://github.com/Azure/DotNetty/issues/498
+        public static readonly ConstantPool Pool = new AttributeConstantPool();
+
+        sealed class AttributeConstantPool : ConstantPool
+        {
+            protected override IConstant NewConstant<TValue>(int id, string name) => new AttributeKey<TValue>(id, name);
+        };
+    }
+
     /// <summary>
     ///     Key which can be used to access <seealso cref="Attribute" /> out of the <see cref="IAttributeMap" />. Be aware that
     ///     it is not be possible to have multiple keys with the same name.
@@ -14,12 +26,7 @@ namespace DotNetty.Common.Utilities
     /// </typeparam>
     public sealed class AttributeKey<T> : AbstractConstant<AttributeKey<T>>
     {
-        public static readonly ConstantPool Pool = new AttributeConstantPool();
-
-        sealed class AttributeConstantPool : ConstantPool
-        {
-            protected override IConstant NewConstant<TValue>(int id, string name) => new AttributeKey<TValue>(id, name);
-        };
+        public static readonly ConstantPool Pool = AttributeKey.Pool;
 
         /// <summary>Returns the singleton instance of the {@link AttributeKey} which has the specified <c>name</c>.</summary>
         public static AttributeKey<T> ValueOf(string name) => (AttributeKey<T>)Pool.ValueOf<T>(name);


yyjdelete:
Some test failed with the below message. Seems just an concurrent issue with `InternalLoggerFactoryTest`, and not related to this PR, will push again to check this. Not sure why `DisableParallelization = true` not works.
 System.TypeInitializationException : The type initializer for 'DotNetty.Common.Utilities.HashedWheelTimer' threw an exception.
<details>

```
---- Moq.MockException : ILoggerProvider.CreateLogger("DotNetty.Common.Utilities.HashedWheelTimer") invocation failed with mock behavior Strict.
All invocations on the mock must have a corresponding setup.
Stack Trace:
   at DotNetty.Common.Utilities.HashedWheelTimer..ctor(TimeSpan tickInterval, Int32 ticksPerWheel, Int64 maxPendingTimeouts) in C:\projects\dotnetty\src\DotNetty.Common\Utilities\HashedWheelTimer.cs:line 105
   at DotNetty.Common.Tests.Utilities.HashedWheelTimerTest.TestTimerShouldThrowExceptionAfterShutdownForNewTimeouts() in C:\projects\dotnetty\test\DotNetty.Common.Tests\Utilities\HashedWheelTimerTest.cs:line 86
----- Inner Stack Trace -----
   at Moq.FindAndExecuteMatchingSetup.Handle(Invocation invocation, Mock mock) in C:\projects\moq4\src\Moq\Interception\InterceptionAspects.cs:line 145
   at Moq.Mock.Moq.IInterceptor.Intercept(Invocation invocation) in C:\projects\moq4\src\Moq\Interception\Mock.cs:line 21
   at Castle.DynamicProxy.AbstractInvocation.Proceed()
   at Castle.Proxies.ILoggerProviderProxy.CreateLogger(String categoryName)
   at Microsoft.Extensions.Logging.Logger..ctor(LoggerFactory loggerFactory, String name)
   at Microsoft.Extensions.Logging.LoggerFactory.CreateLogger(String categoryName)
   at DotNetty.Common.Internal.Logging.InternalLoggerFactory.GetInstance(String name) in C:\projects\dotnetty\src\DotNetty.Common\Internal\Logging\InternalLoggerFactory.cs:line 94
   at DotNetty.Common.Utilities.HashedWheelTimer..cctor() in C:\projects\dotnetty\src\DotNetty.Common\Utilities\HashedWheelTimer.cs:line 21
```

</details>

PR #516, opened 2019-10-12
Fix bug memory to increase

waleswood:
Decreases the reference count by 1，If you don't call release, memory will explode


nayato:
There's no need to release the buffer as it is being put back to channel:
- When buffer is received in ChannelRead, it has ref count of 1
- Then it is given to WriteAsync. From that point channel is on point to decrement ref count by 1 once it's done sending the buffer.

PR #522, opened 2019-11-08
Ignore WAKEUP_TASK in SingleThreadEventExecutor.PollTask

yyjdelete:
Without this, it may lead to an infinite loop.

### DIFF
--- a/src/DotNetty.Common/Concurrency/SingleThreadEventExecutor.cs
+++ b/src/DotNetty.Common/Concurrency/SingleThreadEventExecutor.cs
@@ -489,31 +489,36 @@ namespace DotNetty.Common.Concurrency
         {
             Contract.Assert(this.InEventLoop);
 
-            IRunnable task;
-            if (!this.taskQueue.TryDequeue(out task))
+            while (true)
             {
-                this.emptyEvent.Reset();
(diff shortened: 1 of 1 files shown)


yyjdelete:
I didn't find a way to create an test for this. It an concurrency issue and only happen in rare case.

Shutdown will never finished if an task is enqueued during `CleanupAndTerminate()`. If there is any tasks left, `RunAllTasks` will return true and call `WakeUp(true)`, and `WakeUp(true)` will enqueue an `WAKEUP_TASK` and it makes the next call to `RunAllTasks` still return true.
https://github.com/Azure/DotNetty/blob/a9d0723ba5994c1b98bb43ed452996d51beb1a7c/src/DotNetty.Common/Concurrency/SingleThreadEventExecutor.cs#L115-L119
https://github.com/Azure/DotNetty/blob/a9d0723ba5994c1b98bb43ed452996d51beb1a7c/src/DotNetty.Common/Concurrency/SingleThreadEventExecutor.cs#L375-L381
https://github.com/Azure/DotNetty/blob/a9d0723ba5994c1b98bb43ed452996d51beb1a7c/src/DotNetty.Common/Concurrency/SingleThreadEventExecutor.cs#L317-L328

PR #557, opened 2021-05-15
Update to NetStandard 2.0

joebeernink:
Updated DotNetty.Common to .NETStandard 2.0 to allow this code to be integrated with projects targeting newer versions of the Microsoft.Extensions.* nugets and newer versions of Google.Protobuf

Eliminated Support for .NET45
Eliminated Support for .NETStandard1.3
Updated all nugets consumed to latest version

Ran All Unit Tests:
image

### DIFF
--- a/examples/Discard.Client/Discard.Client.csproj
+++ b/examples/Discard.Client/Discard.Client.csproj
@@ -1,12 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp2.0;net451</TargetFrameworks>
-    <NetStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">1.6.1</NetStandardImplicitPackageVersion>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <NetStandardImplicitPackageVersion>2.0.3</NetStandardImplicitPackageVersion>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net451' ">
-    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
   </PropertyGroup>
   <ItemGroup>
     <None Include="..\..\shared\dotnetty.com.pfx" Link="dotnetty.com.pfx">
--- a/src/DotNetty.Codecs/DotNetty.Codecs.csproj
+++ b/src/DotNetty.Codecs/DotNetty.Codecs.csproj
@@ -1,13 +1,13 @@
 ﻿<?xml version="1.0" encoding="utf-8"?><Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.3;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <IsPackable>true</IsPackable>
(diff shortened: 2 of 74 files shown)


src/DotNetty.Codecs/DotNetty.Codecs.csproj
```diff
@@ -1,13 +1,13 @@
...
-    <TargetFrameworks>netstandard1.3;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <IsPackable>true</IsPackable>
     <PackageId>DotNetty.Codecs</PackageId>
     <Description>General purpose codecs for DotNetty</Description>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <AssemblyTitle>DotNetty: codecs</AssemblyTitle>
     <NeutralLanguage>en-US</NeutralLanguage>
-    <VersionPrefix>0.6.0</VersionPrefix>
+    <VersionPrefix>2.0.0</VersionPrefix>
```
nayato: let's go for 0.7. For 0.x, incrementing "x" implies breaking changes (which these are). 2.0 implies it is a version with support commitment and I doubt we want to do that.

src/DotNetty.Handlers/Tls/TlsHandler.cs
```diff
@@ -718,7 +711,6 @@ public void ExpandSource(int count)
                 }
                 this.sslOwnedBuffer = default(ArraySegment<byte>);
 
-#if NETSTANDARD1_3
```
nayato, line 721: SslStream API necessary for efficient interaction in synchronous context are present in .net standard 2.0 (unlike in 1.3). So, for TlsHandler.cs specifically, we should use `#else` branch instead now.

src/DotNetty.Transport/Channels/Sockets/AbstractSocketByteChannel.cs
```diff
@@ -153,22 +153,8 @@ public override void FinishRead(SocketChannelAsyncOperation operation)
         protected override void ScheduleSocketRead()
         {
             SocketChannelAsyncOperation operation = this.ReadOperation;
-            bool pending;
-#if NETSTANDARD1_3
-            pending = this.Socket.ReceiveAsync(operation);
-#else
-            if (ExecutionContext.IsFlowSuppressed())
```
nayato, line 160: `ExecutionContext` was not available .net standard 1.3. It is in 2.0 so we should keep `#else` branch here (and elsewhere with `ExecutionContext`) instead.

test/DotNetty.Handlers.Tests/MediationStream.cs
```diff
@@ -39,38 +39,6 @@ public override void SetLength(long value)
 
         public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) => this.writeDataFunc(new ArraySegment<byte>(buffer, offset, count));
 
-#if !NETCOREAPP1_1
```
nayato: pls revert per suggestion in TlsHandler.

examples/Discard.Client/Discard.Client.csproj
```diff
@@ -1,13 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp2.0;net451</TargetFrameworks>
-    <NetStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">1.6.1</NetStandardImplicitPackageVersion>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
```
drwill-ms, line 4: The .NET SDK still needs the older targets too. Can you make this additive only?

PR #583, opened 2022-04-14
Fix infinite loop caused by TLS Handler(Only .NET6)

kenlon:
Fix infinite loop caused by TLS Handler, and the memory exploded instantly.

### DIFF
--- a/src/DotNetty.Handlers/Tls/TlsHandler.cs
+++ b/src/DotNetty.Handlers/Tls/TlsHandler.cs
@@ -429,6 +429,13 @@ namespace DotNetty.Handlers.Tls
                             break;
                         }
                         int read = currentReadFuture.Result;
+                        
+                        if (read == 0)
+                        {
+                            //Stream closed
+                            return;
+                        }
+                        
                         AddBufferToOutput(outputBuffer, read, output);
                     }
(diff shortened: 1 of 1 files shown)


kenlon:
.NET 6.0.202 has avoided (not fixed) this problem, and read operations on closed connections will throw exception, as in versions prior to .NET 6

jianjiuss:
I have same question with this.Can i use .netcore 3.1 to fix this problem?

kenlon:
> I have same question with this.Can i use .netcore 3.1 to fix this problem?

YES, Or update to the latest version .NET 6(>=6.0.202)

jianjiuss:
> I had the same problem with version .NET 6(6.0.5.31213).Now i downgrade to .net core 3.1(3.1.25), It didn't seem to reappear again.

jianjiuss:
>

jianjiuss:
> > I have same question with this.Can i use .netcore 3.1 to fix this problem?
> 
> YES, Or update to the latest version .NET 6(>=6.0.202)

I had the same problem with version .NET 6(6.0.5.31213).Now i downgrade to .net core 3.1(3.1.25), It didn't seem to reappear again.

Aaronontheweb:
I would be very helpful is this could get merged in. Ran into this issue the other day.

PR #616, opened 2023-12-01
Fix the issue of TLS handshake failure in .NET 8 environment.

nivalxer:
A recent change was made to SslStream which calls Stream.Read on the inner stream with a zero byte buffer. So, in the .NET 8 environment, all TLS sessions will get stuck at the initialization phase.

- [x] Fix the issue of TLS handshake failure in .NET 8 environment
- [x] Add .NET 8 as the target for all test case projects.

~~AppVeyor does not currently support the .NET 8 SDK.~~

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 version: 0.7.0.{build}
 pull_requests:
   do_not_increment_build_number: true
-image: Visual Studio 2019
+image: Visual Studio 2022
 build_script:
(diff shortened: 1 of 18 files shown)


nivalxer:
@microsoft-github-policy-service agree

marcschier (approved):

nivalxer:
I have been running in a production environment with TLS encrypted communication for 1 month in .NET 8 without any issues. I hope we can merge it as soon as possible.

PR #629, opened 2024-09-21
Add GetGZipHeader method to provide OS-specific Gzip headers for unit…

ikpil:
Add GetGZipHeader method to provide OS-specific Gzip headers for unit testing based on the runtime platform.


ikpil:
@microsoft-github-policy-service agree

